PR #48, opened 2018-04-26
Fix usage of decodeLogs

ajsantander:
Adapts to zos-lib PR #73 also named fix-decodeLogs-1
https://github.com/zeppelinos/zos-lib/pull/73


frangio:
Not sure why the changed line would be causing the tests to fail. Doesn't the code basically do the same as before?

PR #49, opened 2018-04-26
Considers looking for a stdlib address injected with deploy-all

### LINKED ISSUE #50: Using deploy-all injects a stdlib but sync can't find it's address
When using deploy-all, a stdlib's contracts are deployed in a directory and this directory's address is written into the package.zos.network.json file.

However, if he sync command is ran afterwards, the stdlib's deployed address is sought for within the node_modules package, ignoring the injected version.

As a result, using deploy-all is not really usable as is.

ajsantander:
Fixes #50

### DIFF
--- a/src/models/Stdlib.js
+++ b/src/models/Stdlib.js
@@ -54,8 +54,15 @@ export default class Stdlib {
 
   getDeployed(network) {
     if (!network) throw "Must specify network to read stdlib deployment address";
-    const networkInfo = JSON.parse(fs.readFileSync(`node_modules/${this.name}/package.zos.${network}.json`));
-    return networkInfo.provider.address;
+    let networkInfo;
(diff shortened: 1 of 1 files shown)


facuspagnuolo:
I think we should prioritize local-deployed stdlibs instead, do you agree?
I'm saying just based on how `sync` works right now (see [here](https://github.com/zeppelinos/zos-cli/blob/b1d82a1803e61e4975adef6eb9a746dc39a4b6b9/src/scripts/sync.js#L39-L41))

frangio:
Instead of using `try ... catch` blocks, I think this could be implemented by checking if the files exist first.

spalladino:
See #58

spalladino:
Implemented as #58

PR #57, opened 2018-04-27
Build one entry point for each command

### LINKED ISSUE #6: Use commander to define commands arguments and generate documentation
Have each script command register itself as a command in [commander](https://www.npmjs.com/package/commander), so it's possible to validate each argument and easily provide documentation.

facuspagnuolo:
**MERGE AFTER #55** (change base branch to the [refactor branch](https://github.com/zeppelinos/zos-cli/tree/refactor_branch) before merging)
Fixes #6 
Related to #9

This PR introduces a new entry point for each command. It allows us to use truffle just when needed and build custom helpers for each command.

### DIFF
--- a/bin/developers.js
+++ b/bin/developers.js
@@ -2,7 +2,7 @@ const initDistribution = require('../src/commands/developers/init-distribution')
 const deploy = require('../src/commands/developers/deploy')
 const register = require('../src/commands/developers/register')
 
-module.exports = function processDevelopersCommands(program) {
+module.exports = function registerDevelopersCommands(program) {
   initDistribution(program)
   deploy(program)
   register(program)
--- a/bin/developers.js
+++ b/bin/developers.js
@@ -1,0 +1,9 @@
+const initDistribution = require('../src/commands/developers/init-distribution')
+const deploy = require('../src/commands/developers/deploy')
+const register = require('../src/commands/developers/register')
+
+module.exports = function processDevelopersCommands(program) {
+  initDistribution(program)
+  deploy(program)
+  register(program)
+}
--- a/bin/zos-cli.js
+++ b/bin/zos-cli.js
@@ -1,15 +1,15 @@
 #! /usr/bin/env node
 
 const program = require('commander')
-const processUserCommands = require('./users')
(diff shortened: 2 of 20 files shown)


bin/developers.js
```diff
@@ -0,0 +1,9 @@
+const initDistribution = require('../src/commands/developers/init-distribution')
+const deploy = require('../src/commands/developers/deploy')
+const register = require('../src/commands/developers/register')
+
+module.exports = function processDevelopersCommands(program) {
```
spalladino: I'd rename `process` to `register`, but it's a detail

src/commands/developers/init-distribution.js
```diff
@@ -0,0 +1,13 @@
+const initDistribution = require('../../scripts/init-distribution')
+
+module.exports = function(program) {
+  program
+    .command('init-distribution <name> <kernel>')
+    .description("Initialize your distribution project for standard libraries .\n  " +
+      "Provide a distribution <name> for your project\n  " +
+      "Provide the zeppelin_os <kernel> address where your releases are going to be published")
```
spalladino: Remember that backticks allow for [multiline strings](https://stackoverflow.com/a/805113/12791)

bin/zos-cli.js
```diff
@@ -1,15 +1,15 @@
 #! /usr/bin/env node
 
-const Path = require('path')
-const shelljs = require('shelljs')
-
-function run() {
-  const bin = 'npx truffle exec'
-  const script = Path.resolve(__dirname, '../src/entrypoint.js')
```
spalladino: We should be able to remove the `sed` script from the prepackage script in package.json after this change, right?

bin/developers.js
facuspagnuolo: agree!

bin/zos-cli.js
facuspagnuolo: definitely!

facuspagnuolo: I addressed this comment here https://github.com/zeppelinos/zos-cli/pull/63

PR #58, opened 2018-04-27
Fix sync after deploy all with stdlib

### LINKED ISSUE #30: Sync command to development fails if there is a stdlib registered
An initial deploy-all command in development syncs the app manager and deploys an entire stdlib for testing. Subsequent sync commands should make use of the pre-deployed address, instead of failing due to not finding the `package.zos.development.json` file in the dependency.

spalladino:
Fixes #30 #50. Supersedes #49.

Adds a `customDeployed` flag to the `stdlib` struct in `zos.network.json` file to signal that the stdlib is a custom deploy, along with the stdlib name. When calling sync, the command first checks that value, and if the custom deployed stdlib matches the desired lib, uses that address instead of the one provided by the package.

This approach allows invoking sync safely after a deploy-all for testing. It's only improvement over #49 is that it's possible to make a custom deploy of a stdlib to a network where it was already deployed.


src/scripts/sync.js
```diff
@@ -35,11 +38,23 @@ async function sync({ network, from, packageFileName, handleStdlib }) {
...
+  } else if (zosPackage.stdlib) {
+    const networkStdlibInfo = zosNetworkFile.stdlib
+    // If syncing on top of a custom deploy of a stdlib, link to that deploy
+    if (networkStdlibInfo && networkStdlibInfo.customDeploy && networkStdlibInfo.name === zosPackage.stdlib.name) {
+      log.info("Using existing custom deployment of stdlib")
+      await appManager.setStdlib(networkStdlibInfo.address)
+    // Otherwise, link to the public deployed version
+    } else {
+      log.info("Connecting to public deployment of stdlib")
+      const stdlibAddress = await appManager.setStdlib(zosPackage.stdlib)
```
ajsantander: You might want to do this only if needed, i.e. if the stdlib is not set or if it's different than what is being set.

test/helpers/assertions.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function (chai, utils) {
```
ajsantander: There is a new helper being assimilated in a PR:
```zos-lib/lib/validateAddress.js```

ajsantander:
Had a look :+1: 
Just a couple of minor comments above, which are totally ignorable.

test/helpers/assertions.js
spalladino: We can do this later once it's merged and we have that helper available

spalladino:
@ajsantander done! Please review the new commit, and feel free to merge directly afterwards.

PR #59, opened 2018-04-28
Move contracts artifacts requirements to objects/functions context

facuspagnuolo:
**MERGE AFTER #57** (change base branch to the [refactor branch](https://github.com/zeppelinos/zos-cli/tree/refactor_branch) before merging)

Currently we are requiring contracts artifacts at the top level of the JS files, instead of doing it just within thee context they are required (e.g. within a function, inside a constructor, etc). This means we couldn't require such files before we have contracts provision solved. This PR moves all the artifact requirements to the context they are needed.

Additionally, I turned the `makeContract` function into a new JS object called `ContractsProvider`. I know it may overlap the `ContractProvider` we used in `zos-lib`, but that is what actually does within the `zos-cli` context. Please feel free to jump in with alternatives tho.


facuspagnuolo:
I like the idea but I don't see how modules will trigger an object getter, we can work out a solution tho

PR #62, opened 2018-05-02
Split kernel model into a wrapper and a provider

facuspagnuolo:
**MERGE AFTER #61** (change base branch to the [refactor branch](https://github.com/zeppelinos/zos-cli/tree/refactor_branch) before merging)

Following what I described in #60, this PR splits the kernel model into a wrapper and a provider.
Note there is no need of a deployer here, however we have that code spread somehow within the kernel project.


facuspagnuolo:
done @spalladino , please review

PR #67, opened 2018-05-03
Model a file system wrapper

facuspagnuolo:
**MERGE AFTER #66** (change base branch to the [refactor branch](https://github.com/zeppelinos/zos-cli/tree/refactor_branch) before merging)

This PR adds a new FileSystem model wrapping fs module functionality. It allows us to read, parse and write files more friendly.

### DIFF
--- a/src/models/FileSystem.js
+++ b/src/models/FileSystem.js
@@ -1,0 +1,20 @@
+import fs from 'fs'
+
+export default {
+  read(filename) {
+    return fs.readFileSync(filename)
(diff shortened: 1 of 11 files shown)


spalladino:
-1 to have this file in `models`, it should be part of `utils`, since it is not an app model but a set of helpers for interacting with json files in the filesystem.

facuspagnuolo:
+1 to your -1 :p, will do it on the refactor branch directly

PR #68, opened 2018-05-03
Remove testing circumvents from models

facuspagnuolo:
**MERGE AFTER #67** (change base branch to the [refactor branch](https://github.com/zeppelinos/zos-cli/tree/refactor_branch) before merging)

This PR removes all the testing circumvents from models.
The way I did this is to provide a general testing setup function. Said function is in charge of stubbing, handling logs, etc.


spalladino:
Nice! I really like moving test-logic outside of the models themselves. However, I'm not sure I like overwriting the `contract` main function for testing. 

How about moving the stubbing logic to a file required as a [--file](https://mochajs.org/#--file-file) in [mocha.opts](https://mochajs.org/#mochaopts)? I believe it should work regardless of truffle.

PR #72, opened 2018-05-03
Middle layer

### LINKED ISSUE #1: Add a middle scriptable layer to the CLI
Current work in the CLI involves a **"top"** layer that handles json data files that keep track of the project state and a **"bottom"** layer that communicates with the AppManager (and friends) in the blockchain. In this way, the CLI is able to manage a project both by deploying and manipulating contracts in the blockchain and by keeping track of them in local json files. Such management is done by using a set of loose commands in the console. 

It's a good thing that such commands are standalone, but if a user wanted to create a script that manages a project, she would have to either interface with the "top" layer, which means calling a bunch of external scripts, or interface with the "bottom" layer which means manually manipulating the json files that keep track of the project state off-chain.

I suggest that we add a **"middle"** layer that can both handle the json files and communicate with the "bottom" layer that interfaces with the contracts in the blockchain. Like so, the "top" layer will communicate with this "middle" layer instead of the "bottom" layer directly. A project would be manageable both by the CLI and by scripts, without forcing the user to understand the mechanics of the json files or the entire upgradeaility model.

spalladino:
Adds a new AppController class that manages the package.zos file, and a NetworkController class that manages the package.zos.network file and delegates on-chain actions to lib's AppManagerWrapper. The goal is to reduce the complexity of the scripts, and have classes that manage the overall behaviour.

Fixes #1


src/models/AppController.js
```diff
@@ -0,0 +1,86 @@
+import _ from 'lodash';
+import { Logger } from 'zos-lib';
+import { parseJsonIfExists, writeJson } from "zos-lib/lib/utils/FileSystem";
+import { ContractsProvider, AppManagerProvider, AppManagerDeployer } from "zos-lib";
```
facuspagnuolo: `ContractsProvider` is a global object now

src/models/AppController.js
```diff
@@ -0,0 +1,86 @@
...
+import Stdlib from './stdlib/Stdlib';
+import NetworkAppController from './NetworkAppController';
+
+const log = new Logger('AppController');
+
+const DEFAULT_VERSION = '0.1.0';
+
+export default class AppController {
+  constructor(packageFileName) {
+    this.packageFileName = packageFileName || 'package.zos.json';
```
facuspagnuolo: what do you think about moving the default to the constructor declaration? I mean
```js
constructor(packageFileName = 'package.zos.json') { ...
```

src/models/AppController.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  onNetwork(network, txParams, networkFileName) {
+    return new NetworkAppController(this, network, txParams, networkFileName);
+  }
+
+  init(name, version) {
+    if (this.package.name) {
+      throw new Error(`Cannot initialize already initialized package ${this.package.name}`)
```
facuspagnuolo: is it the same as doing `throw 'bla'`?

src/models/AppController.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+    
+    this.package.name = name;
+    this.package.version = version || DEFAULT_VERSION;
```
facuspagnuolo: same here, what about making this clear in the method declaration?

src/models/NetworkAppController.js
```diff
@@ -0,0 +1,170 @@
...
+      throw new Error(`Cannot create network file name from ${appController.packageFileName}`);
+    }
+  }
+
+  async sync() {
+    await this.initApp()
+    await this.syncVersion()
+    await this.fetchProvider()
+    await this.uploadContracts()
+    await this.setStdlib()
```
facuspagnuolo: ❤️

src/utils/jsonFile.js
```diff
@@ -0,0 +1,19 @@
...
+  fs.writeFileSync(fileName, data)
+}
+
+export function exists(fileName) {
+  return fs.existsSync(fileName)
+}
+
+export function tryReadFrom(fileName) {
+  return exists(fileName) ? readFrom(fileName) : null;
+}
```
facuspagnuolo: we can remove this file now

test/scripts/create-proxy.test.js
```diff
@@ -1,9 +1,9 @@
 import init from "../../src/scripts/init.js";
 import sync from "../../src/scripts/sync.js";
 import { cleanup, cleanupfn } from "../helpers/cleanup.js";
+import { parseJson } from "zos-lib/lib/utils/FileSystem";
```
facuspagnuolo: you will need to fix these filesystem imports

src/models/AppController.js
```diff
@@ -0,0 +1,86 @@
...
+
+    // TODO: Do not clean up contracts listing and stdlib, inherit from previous version
+    this.package.contracts = {};
+    delete this.package['stdlib'];
+  }
+
+  setStdlib(stdlibNameVersion, installDeps = false) {
+    if (stdlibNameVersion) {
+      const stdlib = new Stdlib(stdlibNameVersion);
+      if (installDeps) stdlib.install();
```
facuspagnuolo: missing await

src/models/stdlib/Stdlib.js
```diff
@@ -9,13 +10,25 @@ export default class Stdlib {
...
 
+  async getContract(contractAlias) {
+    const implementationName = this.jsonData.contracts[contractAlias]
+    if (!implementationName) throw `Contract ${contractAlias} not found in package`
+    const contractData = fs.parseJson(`node_modules/${name}/build/contracts/${implementationName}.json`)
+    return ContractsProvider.getByJSONData(contractData)
+  }
+  
+  async install() {
+    return StdlibInstaller.call(this.nameAndVersion)
```
facuspagnuolo: I think we shouldn't return here

src/models/AppController.js
```diff
@@ -0,0 +1,86 @@
...
+    return this._package;
+  }
+
+  async getContractClass(contractAlias) {
+    const contractName = this.package.contracts[contractAlias];
+    if (contractName) {
+      return ContractsProvider.getFromArtifacts(contractName);
+    } else if (this.hasStdlib()) {
+      const stdlib = new Stdlib(this.package.stdlib.name);
+      return await stdlib.getContract(contractAlias);
```
facuspagnuolo: What do you think about defining a `getFromStdlib` method in `ContractsProvider`. IMO it will improve two things:
1. no need to instantiate an Stdlib here
2. no need to repeat `getContract` logic in `Stdlib` and `StdlibDeployer`
3. it make sense to keep all providing-contracts logic in a single object

src/models/AppController.js
spalladino: I understand that throwing an instance Error is cleaner, but I cannot find any reliable reference now.

src/models/AppController.js
spalladino: Actually, this caused a test to fail, as setting `null` as the version does not trigger the default value. I'm leaving this as is now.

PR #74, opened 2018-05-07
Remove deploy-all command in favour of flag in sync

### LINKED ISSUE #41: Replace deploy-all with a parameter to sync
Remove the `deploy-all` subcommand and add a parameter `--with-stdlib` to `sync`.

The idea is to discourage use of `deploy-all`, as it should only be used for testing.

spalladino:
Fixes #41


src/commands/users/sync.js
```diff
@@ -8,8 +8,9 @@ module.exports = function(program) {
     .usage('--network <network> [options]')
     .option('-f, --from <from>', 'Set the transactions sender')
     .option('-n, --network <network>', 'Provide a network to be used')
+    .option('--deploy-stdlib', 'Deploys a copy of the stdlib (if any) instead of using the one already published to the network by its author (useful in local testrpc networks)')
```
facuspagnuolo: can we add a shortcut like `-ds` or so? it helps a lot :)

spalladino: I'd agree for frequently used options, but this is not the case. The deploy-stdlib should be used only once (when you set-stdlib). And besides, the goal of the issue was to prevent deploy-all from being over-used. As a general rule, I don't like using too many single-char flags, as they can make a command unreadable.

spalladino: Just my 2 cents though, I'm ok with adding it if it gets more votes. Just remember that shortcuts should be single-letter.

PR #79, opened 2018-05-07
Upgrade multiple proxies with a single command

### LINKED ISSUE #33: Improve upgrade-proxy command
Allow to upgrade all proxies of a given contract, instead of individually specifying addresses, or all proxies altogether. Should be smart enough to avoid upgrading those that have the same implementation.

spalladino:
Command upgrade-proxy will now upgrade all proxies for a given
contract alias if no address is specified. Also, if no contract
alias is set, and a --all flag is set, all contracts in the app
will be upgraded to the new version.

Fixes #33

### DIFF
--- a/test/scripts/upgrade-proxy.test.js
+++ b/test/scripts/upgrade-proxy.test.js
@@ -7,8 +7,13 @@ import upgradeProxy from "../../src/scripts/upgrade-proxy.js";
 import { FileSystem as fs } from "zos-lib";
 import { cleanup, cleanupfn } from "../helpers/cleanup.js";
 
+const Proxy = artifacts.require('Proxy');
+const ImplV1 = artifacts.require('ImplV1');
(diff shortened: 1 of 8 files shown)


test/scripts/upgrade-proxy.test.js
```diff
@@ -7,19 +7,22 @@ import upgradeProxy from "../../src/scripts/upgrade-proxy.js";
...
+const ImplV1 = artifacts.require('ImplV1');
+const ImplV2 = artifacts.require('ImplV2');
+
 const should = require('chai')
       .use(require('chai-as-promised'))
+      .use(require('../helpers/assertions'))
       .should();
 
-contract('upgrade-proxy command', function([_, owner]) {
+contract.only('upgrade-proxy command', function([_, owner]) {
```
facuspagnuolo: ☝️

PR #81, opened 2018-05-08
Check modifications in contracts before add/upgrade proxy

### LINKED ISSUE #22: Remind the user to sync the current package state before managing proxies
If a user runs an `add-implementation` and attempts to create a proxy for it **before** running `sync`, the CLI will fail with a not-found error. Or, even worse, since the backing implementation is the same as it was not upgraded, the [proxy itself](https://github.com/zeppelinos/zos-lib/blob/master/contracts/upgradeability/UpgradeabilityProxy.sol#L59) will fail to upgrade with a VM revert.

Remind the user to sync the current package state to the blockchain before creating or upgrading a proxy.

spalladino:
This PR adds to the network package.zos file the bytecode of the
contract at the time it was deployed. It is then compared to the
bytecode of the contract locally before creating or upgrading a
proxy. If there was a change, then the command fails; this can be
overridden via a --force flag to either command.

Fixes #22.

### DIFF
--- a/src/models/NetworkAppController.js
+++ b/src/models/NetworkAppController.js
@@ -1,6 +1,7 @@
 import _ from 'lodash';
 import { Logger } from 'zos-lib';
 import { FileSystem as fs, AppManagerProvider, AppManagerDeployer } from "zos-lib";
+import { bytecodeDigest } from '../utils/digest';
 import StdlibProvider from './stdlib/StdlibProvider';
 import StdlibDeployer from './stdlib/StdlibDeployer';
 import Stdlib from './stdlib/Stdlib';
@@ -177,7 +178,7 @@ export default class NetworkAppController {
       log.info(`Uploaded ${contractName} at ${contractInstance.address}`);
       this.networkPackage.contracts[contractAlias] = {
         address: contractInstance.address,
-        bytecode: contractClass.bytecode
+        bytecodeHash: bytecodeDigest(contractClass.bytecode)
       };
     }));
   }
@@ -244,8 +245,8 @@ export default class NetworkAppController {
     if (!this.isApplicationContract(contractAlias)) return false;
     if (!this.isContractDeployed(contractAlias)) return true;
     const contractClass = ContractsProvider.getFromArtifacts(contractName);
-    const currentBytecode = contractClass.bytecode;
-    const deployedBytecode = this.networkPackage.contracts[contractAlias].bytecode;
+    const currentBytecode = bytecodeDigest(contractClass.bytecode);
+    const deployedBytecode = this.networkPackage.contracts[contractAlias].bytecodeHash;
     return currentBytecode != deployedBytecode;
   }
 
--- a/src/models/NetworkAppController.js
+++ b/src/models/NetworkAppController.js
@@ -128,13 +128,17 @@ export default class NetworkAppController {
     return this._networkPackage;
   }
 
+  get appAddress() {
+    return this.networkPackage.app && this.networkPackage.app.address;
+  }
+
   writeNetworkPackage() {
     fs.writeJson(this.networkFileName, this.networkPackage);
     log.info(`Successfully written ${this.networkFileName}`)
   }
 
   async initApp() {
-    const address = this.networkPackage.app && this.networkPackage.app.address;
+    const address = this.appAddress;
     this.appManagerWrapper = address
       ? await AppManagerProvider.from(address, this.txParams)
       : await AppManagerDeployer.call(this.package.version, this.txParams);
@@ -142,7 +146,7 @@ export default class NetworkAppController {
   }
 
   async loadApp() {
-    const address = this.networkPackage.app && this.networkPackage.app.address;
+    const address = this.appAddress;
     if (!address) throw new Error("Must deploy app to network");
     this.appManagerWrapper = await AppManagerProvider.from(address, this.txParams);
   }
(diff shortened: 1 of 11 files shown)


spalladino:
Note that this requires the contracts to *have been compiled* in order to detect any changes. Analyzing the source code file does not work, since a contract may inherit from a contract imported from another file; so even if the source of the file stays the same, the resulting bytecode may change.

src/models/NetworkAppController.js
```diff
@@ -132,13 +136,16 @@ export default class NetworkAppController {
...
+    // TODO: Store the implementation's hash or full source/byte code to avoid unnecessary deployments
     return Promise.all(_.map(this.package.contracts, async (contractName, contractAlias) => {
       const contractClass = ContractsProvider.getFromArtifacts(contractName);
       log.info(`Uploading ${contractName} implementation for ${contractAlias}`);
       const contractInstance = await this.appManagerWrapper.setImplementation(contractClass, contractAlias);
       log.info(`Uploaded ${contractName} at ${contractInstance.address}`);
-      this.networkPackage.contracts[contractAlias] = contractInstance.address;
+      this.networkPackage.contracts[contractAlias] = {
+        address: contractInstance.address,
+        bytecode: contractClass.bytecode
```
facuspagnuolo: what about storing a hash of the bytecode here instead of storing all of it? it will improve the readability of the json file

src/models/NetworkAppController.js
```diff
@@ -152,7 +159,7 @@ export default class NetworkAppController {
     const networkStdlib = this.networkPackage.stdlib;
     const hasNetworkStdlib = !_.isEmpty(networkStdlib);
     const hasCustomDeploy = hasNetworkStdlib && networkStdlib.customDeploy;
-    const customDeployMatches = hasCustomDeploy && networkStdlib.name === this.package.stdlib.name;
+    const customDeployMatches = hasCustomDeploy && networkStdlib.name === this.package.stdlib.name && networkStdlib.version === this.package.stdlib.version;
```
facuspagnuolo: we can extract these behaviors to some functions like `hasSameNameAndVersion` or so

src/models/NetworkAppController.js
```diff
@@ -161,9 +168,69 @@ export default class NetworkAppController {
...
+    return stdlib.hasContract(contractAlias);
+  }
+
+  isContractDefined(contractAlias) {
+    return this.isApplicationContract(contractAlias) || this.isStdlibContract(contractAlias);
+  }
+
+  isContractDeployed(contractAlias) {
+    return !this.isApplicationContract(contractAlias) || !_.isEmpty(this.networkPackage.contracts[contractAlias]);
+  }
```
facuspagnuolo: I'm starting to think that this controller is growing a lot and we may want to split it into a hierarchy in a short future

src/models/NetworkAppController.js
```diff
@@ -161,9 +168,69 @@ export default class NetworkAppController {
...
 
     // TODO: Check that package version matches the requested one
+    // TODO: Do not invoke setStdlib if matches existing one
     log.info(`Connecting to public deployment of ${this.package.stdlib.name} in ${this.network}`);
     const stdlibAddress = StdlibProvider.from(this.package.stdlib.name, this.network);
     await this.appManagerWrapper.setStdlib(stdlibAddress);
     this.networkPackage.stdlib = { address: stdlibAddress, ... this.package.stdlib };
   }
+
+  checkContracts(throwIfFail = false) {
```
facuspagnuolo: I would try to be more explicit here, it is not clear what we are checking to me. Maybe we can rename this method to `checkDeployedContractsHaveNotChanged` or so

src/models/NetworkAppController.js
```diff
@@ -161,9 +168,69 @@ export default class NetworkAppController {
...
+      msg = `Contracts ${contractsMissing.join(', ')} are not deployed.`;
+    } else if (!_.isEmpty(contractsChanged)) {
+      msg = `Contracts ${contractsChanged.join(', ')} have changed since the last deploy.`;
+    }
+
+    if (msg && throwIfFail) throw new Error(msg);
+    else if (msg) log.info(msg);    
+  }
+
+  checkContract(contractAlias, throwIfFail = false) {
```
facuspagnuolo: same here

src/models/stdlib/Stdlib.js
```diff
@@ -22,6 +23,11 @@ export default class Stdlib {
     this._packageJson = fs.parseJson(filename)
     return this._packageJson
   }
+
+  hasContract(alias) {
+    if (!this.getPackage().contracts) return false;
+    return !_.isEmpty(this.getPackage().contracts[alias]);
```
facuspagnuolo: We should modify the Object prototype and stop using lodash 🙏

spalladino: We are going to have a heated discussion I'm afraid. I love lodash, and have a deeply rooted hate for modifying prototypes (probably because of several years of monkeypatching metaclasses in Ruby).

spalladino: How about we switch to Clojure, which includes all lodash-like methods in its stdlib?

src/models/NetworkAppController.js
spalladino: I was in doubt about that. On one hand it does improve readability, on the other it allows you to reconstruct a deployment from scratch _without_ needing to have a separate json file with the bytecode. However, we are not using it for anything for now.

src/models/NetworkAppController.js
spalladino: Agree. But after EthBA :-P

spalladino:
@facuspagnuolo fixed most comments, and changed bytecode to bytecodeHash in a separate commit. If you accept the PR, please rebase and don't squash, so the commit with the bytecode hash change is separate and can be reverted in the future if necessary.

PR #89, opened 2018-05-10
docs: follow conventions on example commands

elopio:
The square brackets usually mean that the arguments are optional, so you would document something like:
`zos sync [--network <network>]` 
if there is a default network and the it is optional.

The upper case is usually used for environment variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ truffle init
 Initialize your project with zeppelin_os. The next command will create a new `package.zos.json` file:
 
 ```sh
-zos init [NAME] [VERSION] --network [NETWORK]
+zos init <name> <version> --network <network>
 ```
 For example:
 ```sh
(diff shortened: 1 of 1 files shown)


elopio:
This might sound like a bad idea at first, but we should base our CLI UX and docs on git. Give it some thought before dismissing the idea... :)
It is a very complex tool, and sometimes throws you errors that require you to dig into stackoverflow for hours. But, it has some very cool features and the man, docs, errors and warnings try to guide you as best as they can even when the task you are trying to accomplish is super complex. This is the tool that all free software projects have in common, so it's constantly improving as we learn how to make a scary terminal UI friendly. And for a cli, it's not ugly, after a while you even start liking it.

spalladino:
Like the change. This is also consistent with the autogenerated help from commander. Merging!
```
create-proxy [options] <alias>             Deploy a new proxy to make your contract upgradeable.
```

PR #91, opened 2018-05-11
Add status command

### LINKED ISSUE #43: Consider adding status subcommand
It would show deployment status. We have to think about it more.

spalladino:
Compares package file with package network file and outputs current
status of the application in that network. Includes:

- App version
- Deployed contracts
- Stdlib
- Proxies created

Fixes #43

### DIFF
--- a/src/scripts/status.js
+++ b/src/scripts/status.js
@@ -1,0 +1,111 @@
+import AppController from '../models/AppController';
+import { Logger } from 'zos-lib';
+import _ from 'lodash';
+
+let log = new Logger('scripts/status');
+
+export default async function status({ network, txParams = {}, packageFileName = undefined, networkFileName = undefined, logger = undefined }) {
+  if (logger) log = logger;
+  const appController = new AppController(packageFileName).onNetwork(network, txParams, networkFileName);
+  log.info(`Application status for network ${network}`);
+
+  if (!(await appInfo(appController))) return;
+  if (!(await versionInfo(appController))) return;
+  await stdlibInfo(appController);
+  await contractsInfo(appController);
+  await proxiesInfo(appController);
+}
+
+async function appInfo(appController) {
+  if (!appController.appAddress) {
+    log.info(`Application is not yet deployed`);
+    return false;
+  }
+
+  await appController.loadApp();
+  log.info(`Application is deployed at ${appController.appAddress}`);
+  log.info(`- Proxy factory is at ${appController.appManagerWrapper.factory.address}`);
+  log.info(`- Package is at ${appController.appManagerWrapper.package.address}`);
+  return true;
+}
+
+async function versionInfo(appController) {
+  const requestedVersion = appController.package.version;
+  const currentVersion = appController.appManagerWrapper.version;
+  if (requestedVersion !== currentVersion) {
+    log.info(`- Deployed version ${currentVersion} is out of date (latest is ${requestedVersion})`)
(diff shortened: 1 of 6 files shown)


src/scripts/status.js
```diff
@@ -0,0 +1,111 @@
+import AppController from '../models/AppController';
+import { Logger } from 'zos-lib';
+import _ from 'lodash';
+
+let log = new Logger('scripts/status');
```
facuspagnuolo: `const`?

src/scripts/status.js
```diff
@@ -0,0 +1,111 @@
...
+    log.info('- No proxies created');
+    return;
+  }
+
+  _.each(proxies, (proxyInfos, alias) => {
+    _.each(proxyInfos, ({ address, version }) => {
+      log.info(`- ${alias} at ${address} version ${version}`);
+    });
+  });
+}
```
facuspagnuolo: we should move all these functions to a `StatusChecker` object or sth like that

src/scripts/status.js
```diff
@@ -0,0 +1,111 @@
...
+async function contractsInfo(appController) {
+  log.info('Application contracts:');
+  if (_.isEmpty(appController.package.contracts)) {
+    log.info(`- No contracts registered`);
+    return;
+  }
+
+  _.each(appController.package.contracts, function (contractName, contractAlias) {
+    const isDeployed = appController.isContractDeployed(contractAlias);
+    const hasChanged = appController.hasContractChanged(contractAlias);
```
facuspagnuolo: 👍

src/scripts/status.js
```diff
@@ -0,0 +1,111 @@
...
+    } else {
+      log.info(`- ${fullName} is deployed and up to date`);
+    }
+  });
+}
+
+async function stdlibInfo(appController) {
+  log.info('Standard library:');
+
+  // REFACTOR: Part of this logic is duplicated from NetworkAppController#setStdilb, consider deduplicating it
```
facuspagnuolo: we can move this logic to a `StatusChecker` or different status checkers like `ProxyStatusChecker`, `StdlibStatusChecker`, etc and reuse them from the controllers

src/utils/digest.js
```diff
@@ -1,5 +1,5 @@
 import crypto from 'crypto';
 
 export function bytecodeDigest(bytecode) {
-  return crypto.createHash('sha256').update(Buffer.from(bytecode, 'hex')).digest('hex');
+  return crypto.createHash('sha256').update(Buffer.from(bytecode.replace(/^0x/, ''), 'hex')).digest('hex');
```
facuspagnuolo: 😂

test/scripts/status.test.js
```diff
@@ -0,0 +1,194 @@
...
+  const contractAlias = "Impl";
+  const anotherContractName = "AnotherImplV1";
+  const anotherContractAlias = "AnotherImpl";
+  const version = "0.1.0";
+  const network = "test";
+  const stdlibNameVersion = 'mock-stdlib@0.1.0';
+  const packageFileName = "test/tmp/package.zos.json";
+  const networkFileName = `test/tmp/package.zos.${network}.json`;
+  
+  const logger = new TestLogger();
```
facuspagnuolo: If we have the status checker we can return a result of the current status and check that result without having to test the logs... in that case, the script file could just output the result through the logs

PR #93, opened 2018-05-11
 Do not re-upload contracts on sync if they have not changed

### LINKED ISSUE #23: Avoid syncing entire package state to the blockchain on every sync call
Consider storing the hash of each deployed source file in the network file, to determine whether it needs to be re-deployed or not. See: https://github.com/zeppelinos/zos-cli/blob/99840e66243331637f77188cbbbae90920ecc1b0/src/scripts/sync.js#L34.

spalladino:
Sync no longer reuploads all contracts. Only those with changes in local disk are uploaded, unless a `--reupload` flag is set.

Fixes #23


src/models/NetworkAppController.js
```diff
@@ -169,18 +170,22 @@ export default class NetworkAppController {
...
+          const contractClass = ContractsProvider.getFromArtifacts(contractName);
+          log.info(`Uploading ${contractName} implementation for ${contractAlias}`);
+          const contractInstance = await this.appManagerWrapper.setImplementation(contractClass, contractAlias);
+          log.info(`Uploaded ${contractName} at ${contractInstance.address}`);
+          this.networkPackage.contracts[contractAlias] = {
+            address: contractInstance.address,
+            bytecodeHash: bytecodeDigest(contractClass.bytecode)
+          };
+        }).value()
+    );
```
facuspagnuolo: let's extract the upload logic for a single contract to another function, do you agree?

test/mocks/packages/package-with-contracts.zos.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "Herbs",
+  "version": "1.1.0",
+  "contracts": {
+    "Impl": "ImplV1",
+    "AnotherImpl": "AnotherImplV1"
+  }
+}
```
facuspagnuolo: can we use `package-with-contracts-version-1` and `package-with-contracts-version-2` ?

test/scripts/sync.test.js
```diff
@@ -1,13 +1,16 @@
-import { FileSystem as fs } from 'zos-lib'
+import { FileSystem as fs, AppManagerProvider } from 'zos-lib'
 import sync from "../../src/scripts/sync.js";
 import { cleanup, cleanupfn } from '../helpers/cleanup';
+import { promisify } from 'util';
```
facuspagnuolo: are we using this one?

spalladino: Not anymore. Cleaning up.

src/models/NetworkAppController.js
spalladino: Sounds good!

test/mocks/packages/package-with-contracts.zos.json
spalladino: Renamed package-with-contracts-2 to package-with-contracts-v2, but left this one without changes, since several tests use it without referring to another version.

PR #96, opened 2018-05-14
Fail early and loudly in commands

martriay:
Adding a few checks in `sync`, `create-proxy` and `upgrade-proxy`.
Without them, it fails in a late stage with misleading stacktraces.


martriay:
👍 
Closing in favor of #97.

PR #97, opened 2018-05-14
Improve error handling

### LINKED ISSUE #4: CLI commands should throw explicit errors
i.e. when a command fails, describe what failed

facuspagnuolo:
Fixes #4

### DIFF
--- a/src/models/ErrorHandler.js
+++ b/src/models/ErrorHandler.js
@@ -1,0 +1,32 @@
+import { Logger } from 'zos-lib'
+
+const log = new Logger('Error')
+
+export default class ErrorHandler {
+  constructor(error, { verbose }) {
+    this.error = error
+    this.verbose = verbose
+  }
+
+  call() {
+    const errorName = this.error.name || 'Undefined'
+    const handlerName = `on${errorName}`
+    this[handlerName]()
+  }
+
+  onError() {
+    log.error(this.error.message)
+    this._printStackIfVerbose()
+    process.exit(1)
+  }
+
+  onUndefined() {
+    log.error('There was an undefined error. Please execute the same command again in verbose mode if necessary.')
+    this._printStackIfVerbose()
+    process.exit(1)
+  }
+
+  _printStackIfVerbose() {
+    if(this.verbose) console.log(this.error.stack)
+  }
+}
--- a/src/models/ErrorHandler.js
+++ b/src/models/ErrorHandler.js
@@ -1,6 +1,7 @@
 import { Logger } from 'zos-lib'
 
 const log = new Logger('Error')
+const GENERIC_ERROR_MESSAGE = 'There was an undefined error. Please execute the same command again in verbose mode if necessary.'
 
 export default class ErrorHandler {
   constructor(error, { verbose }) {
@@ -9,24 +10,13 @@ export default class ErrorHandler {
   }
 
   call() {
-    const errorName = this.error.name || 'Undefined'
-    const handlerName = `on${errorName}`
-    this[handlerName]()
-  }
-
-  onError() {
-    log.error(this.error.message)
-    this._printStackIfVerbose()
-    process.exit(1)
-  }
-
-  onUndefined() {
-    log.error('There was an undefined error. Please execute the same command again in verbose mode if necessary.')
+    const errorMessage = this.error.message || GENERIC_ERROR_MESSAGE
+    log.error(errorMessage)
     this._printStackIfVerbose()
     process.exit(1)
   }
 
   _printStackIfVerbose() {
-    if(this.verbose) console.log(this.error.stack)
+    if(this.verbose) log.error(this.error.stack)
   }
 }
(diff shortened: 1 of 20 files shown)


src/models/ErrorHandler.js
```diff
@@ -0,0 +1,32 @@
...
+export default class ErrorHandler {
+  constructor(error, { verbose }) {
+    this.error = error
+    this.verbose = verbose
+  }
+
+  call() {
+    const errorName = this.error.name || 'Undefined'
+    const handlerName = `on${errorName}`
+    this[handlerName]()
```
spalladino: If the name of the error is not "Error", this is going to fail. I'd consider changing this dynamic dispatch to something more simple: if the error has a `message` property, output it; otherwise, go for the "undefined error" approach.

src/models/ErrorHandler.js
```diff
@@ -0,0 +1,32 @@
+import { Logger } from 'zos-lib'
+
+const log = new Logger('Error')
+
+export default class ErrorHandler {
```
spalladino: I guess we'll always disagree on using objects vs plain functions for this kind of stuff :-P

src/models/ErrorHandler.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  onUndefined() {
+    log.error('There was an undefined error. Please execute the same command again in verbose mode if necessary.')
+    this._printStackIfVerbose()
+    process.exit(1)
+  }
+
+  _printStackIfVerbose() {
+    if(this.verbose) console.log(this.error.stack)
```
spalladino: Why console.log and not log.error?

src/models/NetworkAppController.js
```diff
@@ -267,4 +265,8 @@ export default class NetworkAppController {
   isContractDeployed(contractAlias) {
     return !this.isApplicationContract(contractAlias) || !_.isEmpty(this.networkPackage.contracts[contractAlias]);
   }
-}
\ No newline at end of file
+
+  areSameStdlib(aStdlib, anotherStdlib) {
+    return aStdlib.name === anotherStdlib.name && aStdlib.version === anotherStdlib.version
+  }
```
spalladino: -1 to sneaking unrelated refactors in PRs. It's just a detail though, let's leave this one for now, but just keep it in mind for future changes.

src/scripts/add-implementation.js
```diff
@@ -1,8 +1,7 @@
 import AppController from "../models/AppController";
 
-export default function addImplementation({ contractName, contractAlias, packageFileName = undefined }) {
-  if (contractName === undefined) throw new Error('Must provide a contract name')
-  if (!contractAlias) contractAlias = contractName
+export default function addImplementation({ contractName, contractAlias = contractName, packageFileName = undefined }) {
```
spalladino: Are we sure that this works fine in JS? It doesn't in Python, hence the question.

src/scripts/deploy.js
```diff
@@ -6,8 +6,10 @@ const log = new Logger('deploy')
 
 // TODO: remove version param
 export default async function deploy({ version, network, txParams = {}, packageFileName = undefined}) {
+  if (!version) throw Error('A version name must be provided to initialize a new version.')
```
spalladino: Version `"0"` should be valid, but would fail this check thanks to js magic

src/scripts/new-version.js
```diff
@@ -1,7 +1,8 @@
-import Stdlib from "../models/stdlib/Stdlib"
 import AppController from  '../models/AppController'
 
 export default async function newVersion({ version, stdlibNameVersion = undefined, installDeps = false, packageFileName = undefined }) {
+  if (!version) throw Error('A version name must be provided to initialize a new version.')
```
spalladino: Same here for version '0'

src/models/ErrorHandler.js
facuspagnuolo: We should decide a pattern and follow a single one through all the app

src/scripts/add-implementation.js
facuspagnuolo: yes it works, however I'm pushing a new PR to support adding multiple implementations that changes this script

facuspagnuolo:
@spalladino comments addressed, please review!

spalladino:
@facuspagnuolo added one minor change request (reject empty strings for versions). After fixed, merge!

martriay:
Should we consider turning on/off the logger prefix along with the `--verbose` option?

Without `--verbose`:

```
Current provider is at 0xaf310a9e2f06929a5d14c0821accd75bd46a3897
Uploading Employment implementation for Employment
Setting implementation of Employment in directory...
Implementation set: 0x7f132ebccf22c2faab02a40e2169129360df1c1c
```

With `--verbose`:

```
[NetworkAppController] Current provider is at 0xaf310a9e2f06929a5d14c0821accd75bd46a3897
[NetworkAppController] Uploading Employment implementation for Employment
[AppManager] Setting implementation of Employment in directory...
[AppManager]  Implementation set: 0x7f132ebccf22c2faab02a40e2169129360df1c1c
```

facuspagnuolo:
@martriay nice catch, creating a new issue for that

PR #102, opened 2018-05-14
Support adding multiple implementations

### LINKED ISSUE #75: Support adding multiple contracts in the same add-implementation call
Consider changing add-implementation so it can be invoked with multiple contracts, potentially with a file glob (keep in mind that contract name and file name may not match, and a file may include multiple contracts) or a regex. Even better, consider adding a `--all` option that adds all compiled contracts.

facuspagnuolo:
Fixes #75

### DIFF
--- a/src/commands/users/add-implementation.js
+++ b/src/commands/users/add-implementation.js
@@ -1,18 +1,23 @@
 import addImplementation from '../../scripts/add-implementation'
+import addAllImplementations from '../../scripts/add-all-implementations'
 
 module.exports = function(program) {
   program
-    .command('add-implementation <contractNames...>')
-    .usage('<contractName1[:contractAlias1]> ... <contractNameN[:contractAliasN]>')
+    .command('add-implementation [contractNames...]')
+    .usage('[contractName1[:contractAlias1] ... contractNameN[:contractAliasN]] [options]')
     .description(`Register contract implementations.
-      Provide a list of <contractNames> to be registered.
+      Provide a list of [contractNames...] to be registered.
       Provide an alias to register your contract using the notation <contractName:contractAlias>.
-      If non alias is provided, <contractName> will be used by default.`)
-    .action(function (contractNames) {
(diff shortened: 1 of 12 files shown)


src/models/AppController.js
```diff
@@ -58,20 +58,38 @@ export default class AppController {
     this.package.contracts[contractAlias] = contractName;
   }
 
+  addAllImplementations() {
+    // TODO: use new zos-lib fs functions once release
```
spalladino: Given the new release, we should be able to use them now :-)

src/scripts/add-implementation.js
```diff
@@ -1,11 +1,13 @@
...
 
   const appController = new AppController(packageFileName)
-  appController.validateImplementation(contractName)
-  appController.addImplementation(contractAlias, contractName)
-  appController.writePackage()
+
+  contractsData.forEach(({ name, alias }) => {
+    appController.validateImplementation(name)
+    appController.addImplementation(alias, name)
+    appController.writePackage()
```
spalladino: Let's write the package file after the for-loop, so the command is transactional, and does not make any changes if one contract fails

src/commands/users/add-implementation.js
```diff
@@ -1,13 +1,23 @@
...
-      Provide the name of the contract under <contractName>.
-      Provide an [alias] to register your contract, otherwise <contractName> will be used by default.`)
-    .action(function (contractName, contractAlias) {
-      addImplementation({ contractName, contractAlias })
+    .command('add-implementation [contractNames...]')
+    .usage('[contractName1[:contractAlias1] ... contractNameN[:contractAliasN]] [options]')
+    .description(`Register contract implementations.
+      Provide a list of [contractNames...] to be registered.
+      Provide an alias to register your contract using the notation <contractName:contractAlias>.
+      If non alias is provided, <contractName> will be used by default.`)
```
spalladino: non => no

Issue #39, opened 2018-04-24
Compile contracts before sync


frangio:
The `sync` subcommand deploys contracts to the network. It should compile them beforehand if necessary.

spalladino:
Note that this must assume that the user is using a certain tool for compiling (such as truffle), which may tie us too much to that specific tool.

martriay:
@spalladino I think that making this assumption for the time being it's an acceptable tradeoff in favor of the DX.

spalladino:
Ok, let's move forward then!

martriay:
Addressed in https://github.com/zeppelinos/zos-cli/pull/108.

PR #108, opened 2018-05-14
Compile contracts before sync

martriay:
Addresses https://github.com/zeppelinos/zos-cli/issues/39.

### DIFF
--- a/src/utils/runWithTruffle.js
+++ b/src/utils/runWithTruffle.js
@@ -1,12 +1,16 @@
 import Web3 from 'web3'
 import TruffleConfig from 'truffle-config'
+import TruffleCompile from 'truffle-workflow-compile'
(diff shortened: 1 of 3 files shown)


src/utils/runWithTruffle.js
```diff
@@ -19,6 +23,17 @@ function initTruffle(network) {
   });
 }
 
-export default function runWithTruffle(script, network) {
+function compileWithTruffle() {
+  log.info("Compiling contracts with truffle")
```
facuspagnuolo: I think truffle already outputs this one for you

maraoz:
pls rebase

src/utils/runWithTruffle.js
martriay: I left that one in case truffle fails silently.

PR #128, opened 2018-05-15
Add command to generate html documentation for zos-docs

spalladino:
Uses react templating to generate md files with html to be used
in the zos-docs project. React is not included as a dependency but
as a dev dependency, and the project should not include docs when
packaged.

Run it via `npm run gen-docs`

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,7 +4,8 @@
       "targets": {
         "node": "8.9"
       }
-    }]
+    }],
+    "react"
   ],
   "plugins": ["transform-object-rest-spread"]
 }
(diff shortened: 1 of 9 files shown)


spalladino:
Sample sidebars:
```
{
  "cli-api": {
    "OVERVIEW": [
      "cli_main"
    ],
    "COMMANDS": [
      "cli_init",
      "cli_add-implementation",
      "cli_sync",
      "cli_create-proxy",
      "cli_new-version",
      "cli_upgrade-proxy",
      "cli_set-stdlib",
      "cli_status",
      "cli_init-distribution",
      "cli_deploy"
    ]
  }
}
```

PR #131, opened 2018-05-15
Update zos-lib to version 0.1.15

facuspagnuolo:


test/models/ProjectDeployer.test.js
```diff
@@ -21,13 +19,13 @@ contract('ProjectDeployer', function ([_, owner]) {
   });
 
   it('deploys all contracts', async function() {
-    this.app.appManager.address.should.not.be.null;
+    this.app._app.address.should.not.be.null;
```
spalladino: Should be `this.app.address()`

PR #132, opened 2018-05-15
Internal renames for commands

spalladino:
Rename the classes and methods based on the changes made in #117


src/bin/users.js
```diff
@@ -1,19 +1,19 @@
...
   init(program)
   addImplementation(program)
-  sync(program)
+  push(program)
   createProxy(program)
-  newVersion(program)
+  bumpVersion(program)
   upgradeProxy(program)
-  setStdlib(program)
+  linkStdlib(program)
```
facuspagnuolo: why not using `link`, `bump`, `add`, etc... ?

PR #136, opened 2018-05-16
Add freeze command

### LINKED ISSUE #26: Add command to freeze a version
Support [freezing an AppDirectory](https://github.com/zeppelinos/zos-lib/blob/master/contracts/application/versioning/FreezableContractDirectory.sol#L24) via a CLI command

facuspagnuolo:
Fixes #26


src/commands/freeze.js
```diff
@@ -0,0 +1,16 @@
+import freeze from '../scripts/freeze'
+import runWithTruffle from '../utils/runWithTruffle'
+
+module.exports = function(program) {
+  program
+    .command('freeze')
+    .usage('--network <network> [options]')
+    .description('Freeze current release version of your stdlib project.')
```
spalladino: Freeze can be invoked in app projects as well. Also, let @maraoz know after merging, since he's reviewing all CLI commands help now.

spalladino: It's fine with me if we want to ship it as command for libs only, but let's be sure about it. It would also affect whether this logic is in the Lib or BaseController.

src/models/network/NetworkLibController.js
```diff
@@ -1,15 +1,13 @@
...
-const log = new Logger('NetworkLibController');
-
 export default class NetworkLibController extends NetworkBaseController {
   constructor(appController, network, txParams, networkFileName) {
     super(...arguments);
   }
 
   get defaultNetworkPackage() {
-    return { contracts: {}, lib: true };
+    return { contracts: {}, lib: true, frozen: false };
```
spalladino: Creating a new version should set `frozen` back to false

src/commands/freeze.js
spalladino: Scratch that. I thought that `AppDirectory` extended from `FreezableImplementationDirectory`, but it is not the case. Please disregard these comments!

src/models/network/NetworkLibController.js
```diff
@@ -27,11 +25,24 @@ export default class NetworkLibController extends NetworkBaseController {
     this.package = await Package.fetch(address, this.txParams);
   }
 
+  async push(reupload = false) {
+    const requestedVersion = this.packageData.version;
+    const currentVersion = this.networkPackage.version;
+    await super.push(reupload)
+    if (requestedVersion !== currentVersion) this.networkPackage.frozen = false
```
spalladino: Instead of overriding `push`, include `this.networkPackage.frozen = false` in `newVersion` directly, since it is only called when `requestedVersion !== currentVersion`.

test/scripts/freeze.js
```diff
@@ -0,0 +1,36 @@
+'use strict'
```
spalladino: Filename should be `freeze.test.js`

PR #138, opened 2018-05-16
Return created proxy instance from app controller and script

spalladino:


src/scripts/create-proxy.js
```diff
@@ -5,7 +5,7 @@ export default async function createProxy({ contractAlias, initMethod, initArgs,
 
   const appController = new AppController(packageFileName).onNetwork(network, txParams, networkFileName);
   await appController.checkLocalContractDeployed(contractAlias, !force);
-  await appController.createProxy(contractAlias, initMethod, initArgs);
+  const proxy = await appController.createProxy(contractAlias, initMethod, initArgs);
   appController.writeNetworkPackage();
-}
-
+  return proxy;
+}
```
nventuro: Do we want to enforce newlines at EOF?

spalladino:
We should set up a linter later down the road to tackle these issues

Issue #54, opened 2018-04-27
Move version from deploy script to init-distribution


facuspagnuolo:
Currently, the `deploy` scripts requires a version name to deploy a release. 
It should be picked from the `zos.package.json`.

spalladino:
Fixed via #146

Issue #45, opened 2018-04-24
Review developer experience of the stdlib-developer-side commands


facuspagnuolo:
We should review:
- deploy-all
- init-distribution
- deploy

spalladino:
Fixed via #146

Issue #29, opened 2018-04-21
Remove kernel address from package.zos.json for stdlib releases


spalladino:
File package.zos.json should not have any addresses whatsoever, as it is network-independent. Remove it.

spalladino:
Fixed via #146

PR #146, opened 2018-05-16
New library management scripts in CLI

### LINKED ISSUE #135: Add stdlib commands tests

spalladino:
Add new init-lib command to initialize a library instead of an app.
Libraries have no App, and thus manage no proxies. At json level,
they are identified by a `lib` flag in the root.

To manage them, AppController and NetworkAppController was split:
a base BasePackageController has common functionality, while the
AppController and LibController extend it (same for Networked
versions).

The structure of the network json file was also revised:
- All files include current provider and package addresses
- All files include the current version at root level
- Contracts node is at the root of the json file

Commands init-distribution and deploy were removed, now
`init --lib` is used to initialize a library, and deployment 
is managed via `push` same as in an app.

There is a minor REFACTOR item (that requires changing the API
of App in lib) left as a comment. And the naming in the new 
Controller classes may be a bit confusing: the base class is a
BasePackageController, which has packageData; and the network
classes have a packageController, the networkPackage data, and
may use a Package contract. Suggestions are welcome to fix this
(preferrably in a future PR).

Fixes #135 #54 #45 #29


maraoz:
Fixes #135 #54 #45 #29 :clap: 
Reviewing...

src/commands/developers/init-distribution.js
```diff
@@ -1,13 +0,0 @@
...
-  program
-    .command('init-distribution <name> <kernel>')
-    .usage('<name> <kernel>')
-    .description(`Initialize your distribution project for standard libraries.
-      Provide a distribution <name> for your project.
-      Provide the zeppelin_os <kernel> address where your releases are going to be published.`)
-    .action(function (name, kernelAddress) {
-      initDistribution({ name, kernelAddress })
-    })
-}
```
maraoz: :D

src/commands/users/init.js
```diff
@@ -8,13 +9,22 @@ module.exports = function(program) {
...
     .option('--stdlib <stdlib>', 'Standard library to use')
     .option('--no-install', 'Skip installing stdlib npm dependencies')
     .option('--push <network>', 'Push your changes to the specified network')
     .option('-f, --from <from>', 'Set the transactions sender in case you run with --push')
     .action(async function (name, version, options) {
-      const { stdlib: stdlibNameVersion, install: installDeps } = options
-      await init({ name, version, stdlibNameVersion, installDeps })
-      if(options.push) push.action({ network: options.push, from: options.from })
+      if (options.lib) {
+        if (options.stdlib) throw Error("Cannot set an stdlib in a library project")
```
maraoz: Cannot set *a* stdlib

src/models/network/NetworkAppController.js
```diff
@@ -0,0 +1,159 @@
...
+    if (!this.packageController.hasStdlib()) {
+      delete this.networkPackage['stdlib'];
+      return;
+    }
+    const stdlibAddress = await Stdlib.deploy(this.packageData.stdlib.name, this.txParams);
+    this.networkPackage.stdlib = { address: stdlibAddress, customDeploy: true, ... this.packageData.stdlib };
+  }
+
+  async newVersion(versionName) {
+    // REFACTOR: App should return the newly created directory upon newVersion
```
maraoz: :thinking:

test/scripts/create-proxy.test.js
```diff
@@ -56,6 +57,11 @@ contract('create-proxy command', function([_, owner]) {
     await createProxy({ contractAlias: "NotExists", packageFileName, network, txParams }).should.be.rejectedWith(/not found/);
   });
 
+  it('should refuse to create a proxy for a lib project', async function() {
+    editJson(packageFileName, p => { p.lib = true; });
+    await createProxy({ contractAlias, packageFileName, network, txParams }).should.be.rejectedWith(/lib/);
+  });
+
```
maraoz: why do we want to prevent lib developers to instantiate proxies?

spalladino: Discussed offline

PR #149, opened 2018-05-16
Add build/contracts to version control

spalladino:
I'd appreciate some thoughts here: Given that the built contracts are shipped with the package, it makes sense to commit them to version control as well. However, these are built artifacts that _should_ be reconstructable from the sources (as long as the same compiler is used). 

WDYT? Should we add the folder?


frangio:
>Given that the built contracts are shipped with the package, it makes sense to commit them to version control as well.
This is not true. All projects with a compile step (e.g. using Babel) ship built files in the package which shouldn't be committed to the repo.

What is the benefit of having these in the repo?

spalladino:
> This is not true. All projects with a compile step (e.g. using Babel) ship built files in the package which shouldn't be committed to the repo.

It's a good way of seeing it. I'm closing the PR then, we can reopen it later if needed.

PR #170, opened 2018-05-17
Parse input args as JSON

### LINKED ISSUE #156: zos create cannot receive arrays
The arguments pased to the initialization function in the create command are split with `initArgs.split(",")`, causing arrays (e.g. `[1,2,3]`) to not be correctly parsed.

spalladino:
Use the same strategy as remix in parsing input parameters. Note
that the remix code had a bug in the regex, which is fixed here.

Fixes #156

### DIFF
--- a/src/commands/create-proxy.js
+++ b/src/commands/create-proxy.js
@@ -1,5 +1,6 @@
 import createProxy from '../scripts/create-proxy'
 import runWithTruffle from '../utils/runWithTruffle'
+import { parseArgs } from '../utils/input'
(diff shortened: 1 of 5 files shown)


nventuro:
This needs to be reflected in the docs @fiiiu

fiiiu:
@nventuro, do you mean documenting that no spaces are allowed between params, or something else?

nventuro:
I meant the fact that args are parsed as JSON instead of comma separated values

Issue #141, opened 2018-05-16
Review CLI commands help


spalladino:
Each command in the `commands` registers its description and usage info in the `program` object. This is returned when running `zos --help` or `zos COMMAND --help`, and also exported to the documentation site.

Review those docs.

PR #171, opened 2018-05-17
Improve help 

maraoz:
- rename `--stdlib` option in `bump` command to `--link`. E.g: `zos bump 1.2.0 --link openzeppelin-zos@1.1.0`
- improve help message with differentiated main help with per-command help
- improve all commands and parameters help
- rename `--reupload` option in `push` command to `--reset`.
- make main help prettier

Fixes https://github.com/zeppelinos/zos-cli/issues/141

### DIFF
--- a/src/bin/program.js
+++ b/src/bin/program.js
@@ -31,7 +31,6 @@ program
 const commands = [init, add, push, create, bump, upgrade, link, status, freeze]
 
 commands.forEach((c) => {
-  if (c.setup)
   c.setup(program)
 });
 
--- a/src/bin/program.js
+++ b/src/bin/program.js
@@ -1,36 +1,46 @@
+'use strict';
+
 const program = require('commander')
 const { version } = require('../../package.json')
 const registerErrorHandler = require('./errors')
 const Logger = require('zos-lib').Logger;
 
 const init = require('../commands/init')
-const addImplementation = require('../commands/add-implementation')
+const add = require('../commands/add-implementation')
 const push = require('../commands/push')
-const createProxy = require('../commands/create-proxy')
-const bumpVersion = require('../commands/bump-version')
-const upgradeProxy = require('../commands/upgrade-proxy')
-const linkStdlib = require('../commands/link-stdlib')
-const freeze = require('../commands/freeze')
+const create = require('../commands/create-proxy')
+const bump = require('../commands/bump-version')
+const upgrade = require('../commands/upgrade-proxy')
+const link = require('../commands/link-stdlib')
 const status = require('../commands/status')
+const freeze = require('../commands/freeze')
 
 program
   .name('zos')
   .usage('<command> [options]')
+  .description('where <command> is one of:\n' +
+          '\t add, bump, create, init, link, push, status, upgrade.')
   .version(version, '--version')
-  .option('-v, --verbose', 'Verbose mode. Output errors stacktrace and detailed log.')
-  .option('-s, --silent', 'Silent mode. Do not output anything to stderr.')
+  .option('-v, --verbose', 'verbose mode on: output errors stacktrace and detailed log.')
+  .option('-s, --silent', 'silent mode: no output sent to stderr.')
   .on('option:verbose', () => { Logger.verbose(true); } )
   .on('option:silent', () => { Logger.silent(true); } )
+  .option('-v, --verbose', 'switch verbose mode on')
 
-init(program)
-addImplementation(program)
-push(program)
-createProxy(program)
-bumpVersion(program)
-upgradeProxy(program)
-linkStdlib(program)
-freeze(program)
-status(program)
+
+const commands = [init, add, push, create, bump, upgrade, link, status, freeze]
+
+commands.forEach((c) => {
+  if (c.setup)
+  c.setup(program)
+});
+
+program.on('--help', function(){
+  commands.forEach((c) => {
+    console.log('    '+c.signature+'\t'+c.description)
+  });
+  console.log('');
+});
 
 registerErrorHandler(program)
 
--- a/src/bin/program.js
+++ b/src/bin/program.js
@@ -34,7 +34,7 @@ commands.forEach((c) => c.register(program));
 const maxSig = Math.max(...commands.map(c => c.signature.length))
 program.on('--help', function(){
   commands.forEach((c) => {
-    console.log('    '+chalk.bold(c.signature.padEnd(maxSig))+'\t'+c.description)
+    console.log(`    ${chalk.bold(c.signature.padEnd(maxSig))}\t${c.description}`)
   });
   console.log('');
 });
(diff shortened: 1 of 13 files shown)


src/bin/program.js
```diff
@@ -1,36 +1,45 @@
...
+  c.setup(program)
+});
+
+const maxSig = Math.max(...commands.map(c => c.signature.length))
+program.on('--help', function(){
+  commands.forEach((c) => {
+    console.log('    '+chalk.bold(c.signature.padEnd(maxSig))+'\t'+c.description)
+  });
+  console.log('');
+});
```
spalladino: Fancy!

src/bin/program.js
```diff
@@ -1,36 +1,45 @@
...
+const commands = [init, add, push, create, bump, upgrade, link, status, freeze]
+
+commands.forEach((c) => {
+  c.setup(program)
+});
+
+const maxSig = Math.max(...commands.map(c => c.signature.length))
+program.on('--help', function(){
+  commands.forEach((c) => {
+    console.log('    '+chalk.bold(c.signature.padEnd(maxSig))+'\t'+c.description)
```
spalladino: I prefer interpolation with backticks, I find it more clear than concatenation; but it's a matter of personal preference

src/commands/add-implementation.js
```diff
@@ -1,27 +1,31 @@
...
-        const contractsData = contractNames.map(rawData => {
-          let [ name, alias ] = rawData.split(':')
-          return { name, alias: (alias || name) }
-        })
-        addImplementation({ contractsData })
-      }
-      if(options.push) push.action({ network: options.push, from: options.from })
-    })
+const signature = 'add [contractNames...]'
+const description = 'add contract implementations to your project. Provide a list of contract names'
```
spalladino: Why export signature and description separately? They can be retrieved as part of the `Command` object in the program, like `program.commands[0].usage()`, or `program.commands[0].description()`.

spalladino: Regarding the description itself, should we clarify it must be separated by spaces?

src/commands/create-proxy.js
```diff
@@ -1,28 +1,37 @@
...
+const description = 'deploys a new upgradeable contract instance. Provide the <alias> you added your contract with'
+
+module.exports = {
+  signature, description,
+  setup: function(program) {
+    program
+      .command(signature, {noHelp: true})
+      .usage('<alias> --network <network> [options]')
+      .description(description)
+      .option('--init [function]', `call function after creating contract. If no name is given, 'initialize' will be used`)
```
spalladino: If no _function/function name/value_ is given

src/commands/create-proxy.js
```diff
@@ -1,28 +1,37 @@
...
+      .option('--args <arg1, arg2, ...>', 'provide initialization arguments for your contract if required')
```
spalladino: Should we add clarifications on the format? Right now they are parsed as JSON, following what remix does. Maybe add an example even?

src/commands/push.js
```diff
@@ -20,5 +24,4 @@ function action(options) {
   runWithTruffle(async () => await push({ network, deployStdlib, reupload, txParams }), network, ! skipCompile)
 }
 
-module.exports = registerPush
-module.exports.action = action
+module.exports = { signature, description, setup: registerPush, action}
```
facuspagnuolo: Nit: I'd prefer being consistent with how we export modules, I vote to do this way. I think it's clearer to define variables & functions above and export it at the end of the file.

src/commands/push.js
```diff
@@ -1,16 +1,20 @@
...
+    .description(description)
     .usage('--network <network> [options]')
-    .option('-f, --from <from>', 'Set the transactions sender')
-    .option('-n, --network <network>', 'Provide a network to be used')
-    .option('--skip-compile', 'Skips contract compilation')
-    .option('--deploy-stdlib', 'Deploys a copy of the stdlib (if any) instead of using the one already published to the network by its author (useful in local testrpc networks)')
-    .option('--reupload', 'Reuploads all contracts, regardless of not having been modified')
+    .option('-f, --from <from>', 'specify transaction sender address')
+    .option('-n, --network <network>', 'network to be used')
+    .option('-s, --skip-compile', 'skips contract compilation')
```
spalladino: `-s` is used for silent

src/commands/push.js
```diff
@@ -1,16 +1,20 @@
...
+    .option('-d, --deploy-stdlib', 'deploys a copy of the stdlib for development')
+    .option('-r, --reset', 'redeploys all contract implementations (not only the ones that changed)')
```
spalladino: Also, as a general rule, I'd avoid setting short versions for options that are not that frequent, in order to improve readability of all commands.

src/commands/upgrade-proxy.js
```diff
@@ -1,29 +1,39 @@
...
-    .option('--args <arg1, arg2, ...>', 'Provide initialization arguments for your contract if required')
-    .option('--all', 'Skip the alias option and set --all to upgrade all proxies in the application')
-    .option('-f, --from <from>', 'Set the transactions sender')
-    .option('-n, --network <network>', 'Provide a network to be used')
-    .option('--force', 'Force upgrading the proxy even if contracts have local modifications')
-    .action(function (contractAlias, proxyAddress, options) {
-      let initMethod = options.init
-      if(typeof initMethod === 'boolean') initMethod = 'initialize'
+const signature = 'upgrade [alias] [address]'
+const description = 'upgrade contract to a new implementation. Provide the <alias> you added your contract with. If no [address] is provided, all instances of that contract class will be upgraded.'
```
spalladino: Alias is not required, so it should be listed as `[alias]` instead of `<alias>`

src/commands/push.js
spalladino: I think this is the only required change in the entire review

spalladino:
Good cop / bad cop :policeman: 
screenshot from 2018-05-18 11-57-41

src/commands/create-proxy.js
facuspagnuolo: alejo is going to add an example in the docs BTW

src/commands/add-implementation.js
maraoz: @spalladino I didn't find a way to obtain what I called `signature` (`usage()` is different, and too long for the main help).

src/commands/create-proxy.js
maraoz: +1, let's keep the cli help shorter, and expand on docs

src/commands/push.js
maraoz: dammit `commander` didn't complain

src/commands/add-implementation.js
spalladino: I wonder if we are mis-using `usage` then...? We can review this later though

spalladino:
@maraoz looks good. Fix conflicts and squash!

PR #173, opened 2018-05-18
Remove zos-cli TODOs

facuspagnuolo:
Fixes https://github.com/zeppelinos/labs/issues/92
Fixes https://github.com/zeppelinos/zos-cli/issues/175

~However, the CI will fail since there is no `currentStdlib` method in `App` yet, https://github.com/zeppelinos/zos-lib/pull/133 needs to get merged and released first.~


spalladino:
zeppelinos/zos-lib#133 merged, update so we can merge this one!

src/models/network/NetworkAppController.js
```diff
@@ -139,12 +139,14 @@ export default class NetworkAppController extends NetworkBaseController {
       return;
     }
 
-    // TODO: Check that package version matches the requested one
```
spalladino: Note that this TODO is unrelated to the one below. We can convert it to an issue though.

spalladino: Created #175

test/scripts/link-stdlib.test.js
```diff
@@ -30,5 +30,10 @@ contract('link-stdlib command', function() {
     editJson(packageFileName, p => { p.lib = true; });
     await linkStdlib({ stdlibNameVersion: 'mock-stdlib@1.1.0', packageFileName }).should.be.rejectedWith(/lib/);
   });
-  
+
+  it('should raise an error if requested version of stdlib does not match its package version', async function () {
+    editJson('test/mocks/mock-stdlib/zos.json', p => p.version = '1.2.0')
+    await linkStdlib({ stdlibNameVersion: 'mock-stdlib@1.1.0', packageFileName }).should.be.rejectedWith('Requested version 1.1.0 does not match stdlib package version 1.2.0')
+    editJson('test/mocks/mock-stdlib/zos.json', p => p.version = '1.1.0')
```
spalladino: Note that if this test fails, it will fail to restore the zos.json file back to its correct state, and may be inadvertently commited back (since it is not ignored). Consider adding another mock-stdlib, where its package.json version does not match its zos.json.

src/models/network/NetworkAppController.js
spalladino: I see it was tackled in another commit, great!

spalladino:
Sweet! I think we can also close #175 with this, right?

PR #177, opened 2018-05-18
Update README.md with security section

maraoz:


frangio:
We will eventually want to have a much more detailed disclosure and reward policy. For now I think this is OK.

PR #183, opened 2018-05-18
Args flag implies initialization in create and upgrade proxy

### LINKED ISSUE #182: Option --args should imply --init
In create and upgrade proxy, setting `--args` should act as an implicit `--init`.

spalladino:
When setting --args, if --init is not set, it will still invoke
the initializer with the default init method name. Affects create
and upgrade proxy.

Fixes #182

### DIFF
--- a/src/commands/create-proxy.js
+++ b/src/commands/create-proxy.js
@@ -2,7 +2,7 @@
 
 import createProxy from '../scripts/create-proxy'
 import runWithTruffle from '../utils/runWithTruffle'
-import { parseArgs } from '../utils/input'
+import { parseInit } from '../utils/input'
 
 const signature = 'create <alias>'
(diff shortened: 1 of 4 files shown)


src/commands/create-proxy.js
```diff
@@ -20,13 +20,7 @@ module.exports = {
...
       .option('--force', 'force creation even if contracts have local modifications')
       .action(function (contractAlias, options) {
-        let initMethod = options.init
-        if(typeof initMethod === 'boolean') initMethod = 'initialize'
-
-      let initArgs = options.args
-      if(typeof initArgs === 'string') initArgs = parseArgs(initArgs)
-      else if(typeof initArgs === 'boolean' || initMethod) initArgs = []
-
+        const { initMethod, initArgs } = parseInit(options, 'initialize')
```
facuspagnuolo: 😘

PR #198, opened 2018-05-21
Fix input parsing regexs

facuspagnuolo:
I'm removing look-behinds since it's not supported on Mac :/

### DIFF
--- a/src/utils/input.js
+++ b/src/utils/input.js
@@ -7,8 +7,8 @@ export function parseArgs(args) {
     const MATCH_HEX = /(^|([,[]\s*))(0[xX][0-9a-fA-F]+)/g
     args = args.replace(MATCH_HEX, '$2"$3"') // replace non quoted hex string by quoted hex string
 
-    const MATCH_NUMBERS = /(^|([,[]\s*))(\w+)/g
-    args = args.replace(MATCH_NUMBERS, '$2"$3"') // replace non quoted number by quoted number
+    const MATCH_WORDS = /(^|([,[]\s*))(\w+)/g
+    args = args.replace(MATCH_WORDS, '$2"$3"') // replace non quoted number by quoted number
 
     return JSON.parse('[' + args + ']')
   } catch (e) {
--- a/src/utils/input.js
+++ b/src/utils/input.js
@@ -4,8 +4,12 @@ export function parseArgs(args) {
   // The following is inspired from ethereum/remix (Copyright (c) 2016)
   // https://github.com/ethereum/remix/blob/89f34fc4f35eca0c386379550c300205d35bfae0/remix-lib/src/execution/txFormat.js#L51-L53
   try {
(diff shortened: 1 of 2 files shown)


src/utils/input.js
```diff
@@ -4,8 +4,12 @@ export function parseArgs(args) {
   // The following is inspired from ethereum/remix (Copyright (c) 2016)
   // https://github.com/ethereum/remix/blob/89f34fc4f35eca0c386379550c300205d35bfae0/remix-lib/src/execution/txFormat.js#L51-L53
   try {
-    args = args.replace(/(?<=(,|\[|^)\s*)(\d+)(?=\s*(,|\]|$))/g, '"$2"') // replace non quoted number by quoted number
-    args = args.replace(/(?<=(,|\[|^)\s*)(0[xX][0-9a-fA-F]+)(?=\s*(,|\]|$))/g, '"$2"') // replace non quoted hex string by quoted hex string
+    const MATCH_HEX = /(^|([,[]\s*))(0[xX][0-9a-fA-F]+)/g
+    args = args.replace(MATCH_HEX, '$2"$3"') // replace non quoted hex string by quoted hex string
+
+    const MATCH_NUMBERS = /(^|([,[]\s*))(\w+)/g
+    args = args.replace(MATCH_NUMBERS, '$2"$3"') // replace non quoted number by quoted number
```
spalladino: I like that this would now match strings as well, we are actually adding new features. Before, `foo, bar` would not pass, and now this regex should properly catch them and enclose them in quotes. Do you mind adding a test for that, and rename `MATCH_NUMBERS` to `MATCH_WORDS`?

facuspagnuolo: sure!

facuspagnuolo:
@spalladino please review last commit

src/utils/input.js
```diff
@@ -4,8 +4,12 @@ export function parseArgs(args) {
   // The following is inspired from ethereum/remix (Copyright (c) 2016)
   // https://github.com/ethereum/remix/blob/89f34fc4f35eca0c386379550c300205d35bfae0/remix-lib/src/execution/txFormat.js#L51-L53
   try {
-    args = args.replace(/(?<=(,|\[|^)\s*)(\d+)(?=\s*(,|\]|$))/g, '"$2"') // replace non quoted number by quoted number
-    args = args.replace(/(?<=(,|\[|^)\s*)(0[xX][0-9a-fA-F]+)(?=\s*(,|\]|$))/g, '"$2"') // replace non quoted hex string by quoted hex string
+    const MATCH_HEX = /(^|([,[]\s*))(0[xX][0-9a-fA-F]+)/g
+    args = args.replace(MATCH_HEX, '$2"$3"') // replace non quoted hex string by quoted hex string
+
+    const MATCH_WORDS = /(^|([,[]\s*))(\w+)/g
+    args = args.replace(MATCH_WORDS, '$2"$3"') // replace non quoted number by quoted number
```
spalladino: The comment is outdated, but I think we can fix it later down the road :-)

spalladino:
Just in case, let's add (now or later) a few more tests for the parsing, especially words (quoted and unquoted) with spaces and symbols (other than dashes).

PR #202, opened 2018-05-21
Add truffle initializer

facuspagnuolo:
Fixes https://github.com/zeppelinos/zos-cli/issues/203


facuspagnuolo:
@spalladino I followed the gas limit we define in our `truffle.js` file :/, do you think we should change it?

spalladino:
The new commits look good. As for the gas limit, though `ganache-cli` is starting with 90K, it seems that `truffle develop` starts with over 6M. We should try to find that number and use the same. Though defaulting to 5M sounds reasonable at the moment.

PR #207, opened 2018-05-22
tests: setup and cleanup tmp dir

elopio:
It's a better pattern to set up and tear down the tmp dir on each test.
This change ensures that the dir is always clean, and that tests will not
affect each other.

### DIFF
--- a/test/helpers/cleanup.js
+++ b/test/helpers/cleanup.js
@@ -1,11 +1,33 @@
 import fs from 'fs'
+import path from 'path'
 
 export function cleanup(path) {
   try {
     const stat = fs.statSync(path)
-    if(stat.isDirectory()) fs.rmdirSync(path)
+    if(stat.isDirectory()) {
+      rimraf(path)
+    }
     else fs.unlinkSync(path)
   } catch(e) { /* swallow exception */ }
+}
+
+/**
+ * Remove directory recursively
+ * @param {string} dir_path
+ * @see https://stackoverflow.com/a/42505874/3027390
+ */
+function rimraf(dir_path) {
+  if (fs.existsSync(dir_path)) {
+    fs.readdirSync(dir_path).forEach(function(entry) {
+      var entry_path = path.join(dir_path, entry);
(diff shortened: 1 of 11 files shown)


maraoz:
pls rebase

test/helpers/cleanup.js
```diff
@@ -1,13 +1,35 @@
...
+      var entry_path = path.join(dir_path, entry);
+      if (fs.lstatSync(entry_path).isDirectory()) {
+        rimraf(entry_path);
+      } else {
+        fs.unlinkSync(entry_path);
+      }
+    });
+    fs.rmdirSync(dir_path);
+  }
+}
```
facuspagnuolo: I like this! but we should provide it from `lib`, can you move it there?

test/scripts/status.test.js
```diff
@@ -23,22 +24,20 @@ contract('status command', function([_, owner]) {
...
-  
-  beforeEach('cleanup', async function() {
-    cleanup(packageFileName)
-    cleanup(networkFileName)
+  const tmpDir = "test/tmp";
+  const packageFileName = `${tmpDir}/zos.json`;
+  const networkFileName = `${tmpDir}/zos.${network}.json`;
+
+  beforeEach('setup', async function() {
+    fs.createDir(tmpDir);
```
facuspagnuolo: if we need to do this in all tests we can move it to the `setup.js` file, do you agree?

test/helpers/cleanup.js
elopio: https://github.com/zeppelinos/zos-lib/pull/148

test/scripts/status.test.js
elopio: Yes, I would love to make it common. But I'm not finding any docs about setup.js. How can we move a `beforeEach` and `afterEach` there? And can they return a variable for the tests to use?

facuspagnuolo: we can take care of it later then :/

maraoz:
@elopio pls reopen the PR based on a branch on your fork (removing all branches in the main repo getting ready to release)

PR #210, opened 2018-05-22
Use zos-lib contracts instead of truffle artifacts for testing

facuspagnuolo:
Fixes https://github.com/zeppelinos/zos-cli/issues/209

We don't need to use truffle artifacts for testing, we can use our `Contracts` objects
After this, we don't need to have the `Imports` contract anymore


maraoz:
pls rebase

PR #215, opened 2018-05-22
remove 'implementation' from user-facing texts

maraoz:
And also rename some old-named scripts


src/bin/program.js
```diff
@@ -7,12 +7,12 @@ const Logger = require('zos-lib').Logger;
 const chalk = require('chalk')
 
 const init = require('../commands/init')
-const add = require('../commands/add-implementation')
```
ajsantander: Had discussed doing this and @spalladino, who gave doing this a -1. I prefer it like this tho!

ajsantander: Btw, I'm referring to the fact that command file names are the same to command names.

ajsantander:
LGTM

Issue #230, opened 2018-05-26
Cannot set `ContractName` of undefined


PaulRBerg:
## Expected
Should be able to add the contracts with `zos add ContractName`.

## Behaviour
Get `ContractName` to be added when running the command from above. 

Important to note that this happened after, on the same project, I successfully ran both `add` and `create` on an upgradable smart contract, but I made sure to clear `zos.local.json` and the `build` folder and I re-instantiated ganache.

## Tools
• node 9.3.0
• zos 1.0.0
• osx high sierra 10.13.3

Full log output [here](https://raw.githubusercontent.com/PaulRBerg/stackexchange/master/logs/zos-1.0.0.log).

PaulRBerg:
Found the problem. If the `zos.json` has this:

```
  "stdlib": {
    "name": "openzeppelin-zos",
    "version": "1.9.1"
  }
```

And the user runs `zos add ContractName` or `zos create ContractName` only afterward, the error from the title pops out.

facuspagnuolo:
@PaulRBerg I'm sure if I follow how did you get this error. Would you mind adding a detailed step-by-step guide to reproduce it?

PaulRBerg:
Sure!

1. Run `zos init my-project`
2. Run `zos link openzeppelin-zos`
3. Run `zos add ContractName`
4. If using Ganache, restart it
5. Remove the contract from `zos.json` to try to deploy a completely new version
6. The eponymous error occurs

facuspagnuolo:
@PaulRBerg I think I'm not following what you're trying to do in that scenario:
- why would you want to remove the `zos.json` manually?
- what do you mean by "try to deploy a completely new version"? Would you mind telling exactly which commands are you trying to run?

PaulRBerg:
1. I remove it because sometimes I want to deploy a completely new version of the contract, not an upgraded version (completely changing the variables order to not be forced to keep it to satisfy the VM). Also, I'm normally doing lots of debugging and tests, and currently with `zos` to do that on a new version of the contract you have to start a new project.

2. By "trying a new version" I simply mean changing the variables and the functions of the contract, but keeping the name. As stated above, I don't want to run `zos init` to do this.

facuspagnuolo:
@PaulRBerg, 
I see, but it shouldn't be necessary to drop the `zos` file to deploy a new instance manually. Please reach me through our [slack](https://slack.openzeppelin.org/), I'll be glad to help you.

[...]

PR #232, opened 2018-05-26
Initialize contracts list if missing before adding contracts

PaulRBerg:
Fixes https://github.com/zeppelinos/zos-cli/issues/230

After following the [tutorial](https://docs.zeppelinos.org/docs/setup.html) I managed to stumble upon this issue in 30 mins.

The problem occurred when `stdlib` was set in the `zos.json` file, because the `packageData` object was initialized in `LocalBaseController` with its `contracts` property set to null.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,6 @@ node_modules
 zos*
 build
 lib
+.DS_Store
+.idea
 .node-xmlhttprequest-sync-*
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,4 @@ node_modules
 zos*
 build
 lib
-.DS_Store
-.idea
 .node-xmlhttprequest-sync-*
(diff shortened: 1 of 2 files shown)


src/models/local/LocalBaseController.js
```diff
@@ -45,6 +45,7 @@ export default class LocalBaseController {
       log.error(`Contract ${contractName} has an explicit constructor. Move it to an initializer function to use it with ZeppelinOS.`)
     }
 
+    if (!this.packageData.contracts) this.packageData.contracts = {};
```
facuspagnuolo: We shouldn't be initializing the contracts list here, that's part of the `init` command. Have you initialized your project before running `add`?

.gitignore
```diff
@@ -2,4 +2,6 @@ node_modules
 zos*
 build
 lib
+.DS_Store
+.idea
```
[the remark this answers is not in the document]
PaulRBerg: Removed now!

src/models/local/LocalBaseController.js
PaulRBerg: I have. I added more details in #230.

facuspagnuolo:
Closing in favour of https://github.com/zeppelinos/zos-cli/pull/257

PR #233, opened 2018-05-27
Exit process when the task was successfully

wzalazar:
This PR introduces, 

- adds process.exit when the task was successful, it's important for taking the terminal control, for example when you need to use systems of continuous integration

### DIFF
--- a/src/models/local/LocalBaseController.js
+++ b/src/models/local/LocalBaseController.js
@@ -108,5 +108,6 @@ export default class LocalBaseController {
   writePackage() {
     fs.writeJson(this.packageFileName, this.packageData);
     log.info(`Successfully written ${this.packageFileName}`)
+    process.exit(0)
   }
 }
--- a/src/models/network/NetworkBaseController.js
+++ b/src/models/network/NetworkBaseController.js
@@ -41,6 +41,7 @@ export default class NetworkBaseController {
   writeNetworkPackage() {
     fs.writeJson(this.networkFileName, this.networkPackage);
     log.info(`Successfully written ${this.networkFileName}`)
+    process.exit(0)
   }
 
   async init() {


joecoope:
Otherwise open contracts everywhere..
Sent from my Boost Zume 5 on the Telstra Mobile Network
On 27 May 2018 1:12 PM, Walter Zalazar <notifications@github.com> wrote:


This PR introduces,

  *   add process.exit when the task was successful, it's important for taking the terminal control, for example when you need to use systems of continuous integration

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/zeppelinos/zos-cli/pull/233

Commit Summary

  *   exit process

File Changes

  *   M src/models/local/LocalBaseController.js<https://github.com/zeppelinos/zos-cli/pull/233/files#diff-0> (1)
  *   M src/models/network/NetworkBaseController.js<https://github.com/zeppelinos/zos-cli/pull/233/files#diff-1> (1)

Patch Links:

  *   https://github.com/zeppelinos/zos-cli/pull/233.patch
  *   https://github.com/zeppelinos/zos-cli/pull/233.diff

facuspagnuolo:
Thanks @wzalazar. We should do this for all commands I think, right?

wzalazar:
@facuspagnuolo yes, always if you know that the process ends there you should do

facuspagnuolo:
Agree, let's find a way this work for every command then please. We should handle the `exit(0)` in a way we don't have to repeat it for every command.

wzalazar:
hey @facuspagnuolo, I tried to reproduce this issue and I can't. I did with development Network. Maybe is because you don't pin the version of the packages. I recommend pinning version of the packages, with the goal they have all same dependencies in each environment.

facuspagnuolo:
@wzalazar I’m not sure if I follow, why pinning versions will make every command ending with an `exit(0)`?

wzalazar:
@facuspagnuolo with pinning version I mean that we must the same packages in our `package.json`. if not pinning the version maybe when I installed my environment it installed different versions with different behaviors.

```
    "truffle": "^4.1.6",
    "truffle-config": "^1.0.4",
    "truffle-contract": "^3.0.5",
    "truffle-core": "^4.1.8",
```

to 

```
    "truffle": "4.1.6",
    "truffle-config": "1.0.4",
    "truffle-contract": "3.0.5",
    "truffle-core": "4.1.8",
```

facuspagnuolo:
@wzalazar I still don't see why using fixed versions will make every task to end successfully. Based on [semver](https://docs.npmjs.com/getting-started/semantic-versioning), the caret shouldn't change the behavior of what you're expecting of that dependency, it should include improvements, new features or bug-fixes. If that happen we should report an issue to that dependency. Do you agree?

wzalazar:
@facuspagnuolo  it's my assumption, I created this ticket 2 weeks ago, and the last Friday I tried to check this issue and it didn't happen again. For this reason, I suppose this time it installed different versions.

facuspagnuolo:
Let's close it then, I tried to reproduce it and I couldn't too. Thanks @wzalazar!

Issue #243, opened 2018-06-06
Rename init argument to <project-name>


frangio:
https://github.com/zeppelinos/zos-cli/blob/817ebce0f06db62b8144e54f1a70d52c361d0d42/src/commands/init.js#L15

The argument to `zos init` is called `project`, and can result in the following error message which I think is confusing.

```
$ zos init
  error: missing required argument `project'
```

Maybe we should rename it to `project-name`.

PR #251, opened 2018-06-07
Rename `project` param to `project-name` for init command

facuspagnuolo:
Fixes https://github.com/zeppelinos/zos-cli/issues/243


facuspagnuolo:
Thanks @frangio! 
I have the same concern, I'll open an issue to polish the way we declare commands

PR #255, opened 2018-06-14
Introduce a 'zos session <network>' command.

### LINKED ISSUE #122: Implement a `session` command
Lots of commands require a `--network=name` parameter. For example:
```
zos push --network=ropsten
zos create Foo --network=ropsten
zos create ERC20 --network=ropsten
```
It'd be convenient if you could do:
```
zos session --network=ropsten
zos push
zos create Foo
zos create ERC20
```

pinkiebell:
Fixes #122 

Commands like create, freeze, push, status and upgrade will use
<network> unless overriden.

Example:
zos session ropsten
Using "ropsten" as default network unless overriden.
...
zos push
zos create Foo
zos create ERC20 --network=notropsten

### DIFF
--- a/src/commands/session.js
+++ b/src/commands/session.js
@@ -5,16 +5,16 @@ import { setNetwork as setNetwork } from '../scripts/session';
 const signature = 'session';
 const description = 'by providing --network <network>, commands like create, ' +
                     'freeze, push, status and upgrade will use <network> ' +
-                    'unless overriden. Use --remove to undo.';
+                    'unless overriden. Use --close to undo.';
 
 module.exports = {
   signature, description,
   register: function(program) {
     program
       .command(signature, {noHelp: false})
-      .usage('Either --network <network> or --remove')
+      .usage('Either --network <network> or --close')
       .option('--network <network>')
-      .option('--remove')
+      .option('--close')
       .description(description)
       .action(setNetwork);
   }
--- a/src/commands/session.js
+++ b/src/commands/session.js
@@ -1,24 +1,21 @@
 'use strict';
 
-import { FileSystem as fs } from 'zos-lib'
-import { Logger } from 'zos-lib'
+import { setNetwork as setNetwork } from '../scripts/session';
 
-const signature = 'session <network>'
-const description = 'by providing <network>, commands like create, ' +
+const signature = 'session';
+const description = 'by providing --network <network>, commands like create, ' +
                     'freeze, push, status and upgrade will use <network> ' +
-                    'unless overriden.'
+                    'unless overriden. Use --remove to undo.';
 
 module.exports = {
   signature, description,
   register: function(program) {
     program
-      .command(signature, {noHelp: true})
-      .usage('<network>')
+      .command(signature, {noHelp: false})
+      .usage('Either --network <network> or --remove')
+      .option('--network <network>')
+      .option('--remove')
       .description(description)
-      .action(async function (networkName, options) {
-        fs.write('/tmp/.zos_session', networkName)
-        let log = new Logger('commands/session')
-        log.info('Using "' + networkName + '" as default network unless overriden.')
-      })
+      .action(setNetwork);
   }
 }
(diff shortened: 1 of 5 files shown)


src/utils/runWithTruffle.js
```diff
@@ -1,7 +1,16 @@
 import Truffle from '../models/truffle/Truffle';
+import { FileSystem as fs } from 'zos-lib';
 
 export default async function runWithTruffle(script, network, compile = false) {
   const config = Truffle.config()
+
+  if (!network) {
+    try {
+      network = fs.read('/tmp/.zos_session').toString()
```
facuspagnuolo: what about using just `.zos.session` ?

src/commands/session.js
```diff
@@ -0,0 +1,24 @@
...
+  signature, description,
+  register: function(program) {
+    program
+      .command(signature, {noHelp: true})
+      .usage('<network>')
+      .description(description)
+      .action(async function (networkName, options) {
+        fs.write('/tmp/.zos_session', networkName)
+        let log = new Logger('commands/session')
+        log.info('Using "' + networkName + '" as default network unless overriden.')
```
facuspagnuolo: we should move this logic to a script file within the `scripts` dir like the other commands

src/utils/runWithTruffle.js
pinkiebell: Like /tmp/.zos.session or <current working dir> .zos.session ?

facuspagnuolo: without `tmp` I mean

src/commands/session.js
pinkiebell: sounds good. Quick question for coding style: Some files are with- and some are not with semicolons. What do you prefer?

facuspagnuolo:
I'm thinking that we may want to provide a `close-session` command to remove that file

pinkiebell:
What if we simply do:
zos session remove

if (networkName === 'remove') {
  //remove the file
}

This could clash of course if someone has a network named 'remove'.

src/commands/session.js
facuspagnuolo: Two things here:
1. we haven't decided whether it's a must or not using semicolons
2. we have decided to use backtick string interpolation

facuspagnuolo:
I like it, but to avoid that we can apply the following syntax:
- `zos session` can receive two arguments: `network` or `remove`
- only one of both arguments can be passed
- if `network` is given, we expect a value: name of the network
- if `remove` is given, we expect 0 values

Then will have:
- `zos session --network=X` to start a new session on network X, or
- `zos session --remove` to close any session

What do you think?

pinkiebell:
That sounds good. Will do ;)

src/scripts/session.js
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export function setNetwork({ network = undefined, remove = false }) {
+  if ((!network && !remove) || (network && remove)) {
+    throw Error('Please provide either --network <network> or --remove.')
+  }
+
+  let log = new Logger('scripts/session');
+  if (remove) {
+    // no unlink in FileSystem?
```
pinkiebell: We can just use require('fs').unlink. But I wasn't sure to introduce that or not.

facuspagnuolo: you can use the `FileSystem` object provided by `zos-lib`

src/utils/runWithTruffle.js
```diff
@@ -1,7 +1,13 @@
 import Truffle from '../models/truffle/Truffle';
+import { getNetwork as getSessionNetworkOrUndefined } from '../scripts/session';
 
 export default async function runWithTruffle(script, network, compile = false) {
   const config = Truffle.config()
+
+  if (!network) {
+    network = getSessionNetworkOrUndefined();
+  }
```
facuspagnuolo: I guess we can do `network = network || getSessionNetwork()`

martriay:
If `--remove` is meant to close a session, I'd stick with standard terminology, what do you think of having `--close` instead?

PR #257, opened 2018-06-15
Model zos package and network files as objects

### LINKED ISSUE #99: Model a package file object
We are requesting and using the package.zos.json file as is in multiple parts of the tool. 
My proposal is to work with a JS object that carries all the behavior we are doing manually regarding this file.

### LINKED ISSUE #100: Model a package network file
See #99

facuspagnuolo:
Fixes #99
Fixes #100

### DIFF
--- a/src/models/files/ZosNetworkFile.js
+++ b/src/models/files/ZosNetworkFile.js
@@ -11,7 +11,7 @@ export default class ZosNetworkFile {
     this.network = network
     this.fileName = fileName
 
-    const defaults = this.packageFile.isLib() ? { contracts: {}, lib: true, frozen: false } : { contracts: {}, proxies: {} }
+    const defaults = this.packageFile.isLib ? { contracts: {}, lib: true, frozen: false } : { contracts: {}, proxies: {} }
     this.data = fs.parseJsonIfExists(this.fileName) || defaults
   }
 
@@ -71,6 +71,10 @@ export default class ZosNetworkFile {
     return this.data.contracts
   }
 
+  get isLib() {
+    return this.packageFile.isLib
+  }
+
   proxy(alias) {
     return this.proxies[alias]
   }
@@ -79,11 +83,7 @@ export default class ZosNetworkFile {
     return this.contracts[alias]
   }
 
-  isLib() {
-    return this.packageFile.isLib()
-  }
-
-  hasVersion(version) {
+  isCurrentVersion(version) {
     return this.version === version
   }
 
@@ -96,7 +96,7 @@ export default class ZosNetworkFile {
   }
 
   hasMatchingVersion() {
-    return this.packageFile.hasVersion(this.version)
+    return this.packageFile.isCurrentVersion(this.version)
   }
 
   hasCustomDeploy() {
@@ -115,7 +115,7 @@ export default class ZosNetworkFile {
   hasSameBytecode(alias, klass) {
     const deployedBytecode = this.contract(alias).bytecodeHash
     const currentBytecode = bytecodeDigest(klass.bytecode)
-    return currentBytecode !== deployedBytecode
+    return currentBytecode === deployedBytecode
   }
 
   set version(version) {
@@ -157,12 +157,8 @@ export default class ZosNetworkFile {
     }
   }
 
-  setProxy(alias, value) {
-    this.data.proxies[alias] = value
-  }
-
   addProxy(alias, info) {
-    if (!this.proxy(alias)) this.setProxy(alias, [])
+    if (!this.proxy(alias)) this.data.proxies[alias] = []
     this.data.proxies[alias].push(info)
   }
 
--- a/src/models/files/ZosNetworkFile.js
+++ b/src/models/files/ZosNetworkFile.js
@@ -1,0 +1,173 @@
+import _ from 'lodash'
+import { bytecodeDigest } from '../../utils/digest'
+import { Logger, FileSystem as fs } from 'zos-lib'
+
+const log = new Logger('ZosNetworkFile')
+
+export default class ZosNetworkFile {
+
+  constructor(packageFile, network, fileName) {
+    this.packageFile = packageFile
+    this.network = network
+    this.fileName = fileName
+
+    const defaults = this.packageFile.isLib() ? { contracts: {}, lib: true, frozen: false } : { contracts: {}, proxies: {} }
+    this.data = fs.parseJsonIfExists(this.fileName) || defaults
+  }
+
+  get app() {
+    return this.data.app
+  }
+
+  get appAddress() {
+    return this.app && this.app.address
+  }
+
+  get package() {
+    return this.data.package
+  }
+
+  get packageAddress() {
+    return this.package && this.package.address
+  }
+
+  get provider() {
+    return this.data.provider
+  }
+
+  get providerAddress() {
+    return this.provider && this.provider.address
+  }
+
+  get version() {
+    return this.data.version
+  }
+
+  get frozen() {
+    return this.data.frozen
+  }
+
+  get stdlib() {
+    return this.data.stdlib
+  }
+
+  get stdlibName() {
+    return this.stdlib.name
+  }
+
+  get stdlibVersion() {
+    return this.stdlib.version
+  }
+
+  get stdlibAddress() {
+    return this.stdlib.address
+  }
+
+  get proxies() {
+    return this.data.proxies
+  }
+
+  get contracts() {
+    return this.data.contracts
+  }
+
+  proxy(alias) {
+    return this.proxies[alias]
+  }
+
+  contract(alias) {
+    return this.contracts[alias]
+  }
+
+  isLib() {
+    return this.packageFile.isLib()
+  }
+
+  hasVersion(version) {
+    return this.version === version
+  }
+
+  hasProxies() {
+    return !_.isEmpty(this.proxies)
+  }
+
+  hasContract(alias) {
+    return !_.isEmpty(this.contract(alias))
+  }
+
+  hasMatchingVersion() {
+    return this.packageFile.hasVersion(this.version)
+  }
+
+  hasCustomDeploy() {
+    return this.hasStdlib() && this.stdlib.customDeploy
+  }
+
+  hasMatchingCustomDeploy() {
+    return this.hasCustomDeploy() && this.packageFile.hasStdlib(this.stdlib)
+  }
+
+  hasStdlib(stdlib = undefined) {
+    if(stdlib === undefined) return !_.isEmpty(this.stdlib)
+    return this.stdlib.name === stdlib.name && this.stdlib.version === stdlib.version
+  }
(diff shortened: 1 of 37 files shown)


src/models/files/ZosNetworkFile.js
```diff
@@ -0,0 +1,173 @@
...
+
+  contract(alias) {
+    return this.contracts[alias]
+  }
+
+  isLib() {
+    return this.packageFile.isLib()
+  }
+
+  hasVersion(version) {
```
spalladino: -1 to this name, as an app may have several historical versions; it's just that we are not storing them in the json file now. Maybe `currentVersionIs`?

src/models/files/ZosPackageFile.js
```diff
@@ -0,0 +1,113 @@
...
+  hasName(name) {
+    return this.name === name
+  }
+
+  hasStdlib(stdlib = undefined) {
+    if(stdlib === undefined) return !_.isEmpty(this.stdlib)
+    return this.stdlib.name === stdlib.name && this.stdlib.version === stdlib.version
+  }
+
+  hasVersion(version) {
```
spalladino: Same as in network file

src/models/files/ZosPackageFile.js
```diff
@@ -0,0 +1,113 @@
...
+  constructor(fileName = 'zos.json') {
+    this.fileName = fileName
+    this.data = fs.parseJsonIfExists(this.fileName) || {};
+  }
+
+  exists() {
+    return fs.exists(this.fileName)
+  }
+
+  ifMissingThrow(msg) {
```
spalladino: This is more a subjective comment, but I don't like this kind of methods. I'd rather have a caller do `if (!packageFile.exists()) throw Error(msg)` rather than add an extra method here that clutters the interface.

src/models/files/ZosNetworkFile.js
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  setContract(alias, instance) {
+    this.data.contracts[alias] = {
+      address: instance.address,
+      bytecodeHash: bytecodeDigest(instance.constructor.bytecode)
+    }
+  }
+
+  setProxy(alias, value) {
```
spalladino: What's the purpose of `setProxy`? A client should only call `addProxy`.

src/models/files/ZosPackageFile.js
```diff
@@ -0,0 +1,113 @@
...
+
+  contract(alias) {
+    return this.contracts[alias]
+  }
+
+  isLib() {
+    return !!this.lib
+  }
+
+  hasName(name) {
```
spalladino: Same comment as in `ifThrow...`. Why not just let the caller make the comparison, so we don't need yet another method here?

src/models/files/ZosPackageFile.js
```diff
@@ -0,0 +1,113 @@
...
+
+  get contractAliases() {
+    return Object.keys(this.contracts)
+  }
+
+  contract(alias) {
+    return this.contracts[alias]
+  }
+
+  isLib() {
```
spalladino: Why not a getter, for consistency?

src/models/local/LocalAppController.js
```diff
@@ -22,18 +21,8 @@ export default class LocalAppController extends LocalBaseController {
         ? await StdlibInstaller.call(stdlibNameVersion)
         : new Stdlib(stdlibNameVersion)
 
-      this.packageData.stdlib = {
-        name: stdlib.getName(),
-        version: stdlib.getVersion()
-      }
+      const { name, version } = stdlib
+      this.packageFile.stdlib = { name, version }
```
spalladino: If we are restricting the stdlib object to only have name and version, I'd do that filtering in the `packageFile` object rather than here.

src/models/local/LocalBaseController.js
```diff
@@ -1,40 +1,40 @@
...
+    if (this.packageFile.name && !force) {
+      throw Error(`Cannot initialize already initialized package ${this.packageFile.name}`)
     }
-
-    this.packageData.name = name;
-    this.packageData.version = version || DEFAULT_VERSION;
-    this.packageData.contracts = {};
+    this.packageFile.name = name
+    this.packageFile.version = version || DEFAULT_VERSION
+    this.packageFile.contracts = {}
```
spalladino: Why not replace for a `clearContracts`, and remove the setter altogether?

src/models/local/LocalBaseController.js
```diff
@@ -85,19 +84,12 @@ export default class LocalBaseController {
...
-
   getContractClass(contractAlias) {
-    const contractName = this.packageData.contracts[contractAlias];
+    const contractName = this.packageFile.contract(contractAlias);
     if (contractName) {
       return Contracts.getFromLocal(contractName);
-    } else if (this.hasStdlib()) {
-      const stdlibName = this.packageData.stdlib.name;
+    } else if (this.packageFile.hasStdlib()) {
+      const stdlibName = this.packageFile.stdlibName;
```
spalladino: I never liked Demeter, did I ever tell you? :-P

src/models/files/ZosNetworkFile.js
```diff
@@ -0,0 +1,173 @@
...
+
+  get proxies() {
+    return this.data.proxies
+  }
+
+  get contracts() {
+    return this.data.contracts
+  }
+
+  proxy(alias) {
```
spalladino: `proxy` in singular is misleading, as this returns an array

src/models/network/NetworkAppController.js
```diff
@@ -160,46 +145,35 @@ export default class NetworkAppController extends NetworkBaseController {
...
-    const hasCustomDeploy = hasNetworkStdlib && networkStdlib.customDeploy;
-    const customDeployMatches = hasCustomDeploy && this.areSameStdlib(networkStdlib, this.packageData.stdlib);
-
+    const customDeployMatches = this.networkFile.hasCustomDeploy() && this.packageFile.hasStdlib(this.networkFile.stdlib)
     if (customDeployMatches) {
-      log.info(`Using existing custom deployment of stdlib at ${networkStdlib.address}`);
-      await this.app.setStdlib(networkStdlib.address);
-      return;
+      log.info(`Using existing custom deployment of stdlib at ${this.networkFile.stdlibAddress}`);
+      return this.app.setStdlib(this.networkFile.stdlibAddress);
```
spalladino: `await` missing?

src/models/network/NetworkBaseController.js
```diff
@@ -9,42 +8,31 @@ export default class NetworkBaseController {
...
-      this._networkPackage = fs.parseJsonIfExists(this.networkFileName) || _.cloneDeep(this.defaultNetworkPackage);
-    }
-    return this._networkPackage;
+  get isDeployed() {
+    throw Error("Unimplemented function isDeployed()");
   }
 
-  get defaultNetworkPackage() {
-    return { contracts: {} };
+  isLib() {
```
spalladino: Same as above. `isDeployed` is a getter but `isLib` a function, we should be consistent.

src/models/network/NetworkBaseController.js
```diff
@@ -124,28 +108,22 @@ export default class NetworkBaseController {
...
+    const contractName = this.packageFile.contract(contractAlias);
     const contractClass = Contracts.getFromLocal(contractName);
-    const currentBytecode = bytecodeDigest(contractClass.bytecode);
-    const deployedBytecode = this.networkPackage.contracts[contractAlias].bytecodeHash;
-    return currentBytecode !== deployedBytecode;
-  }
-
-  isApplicationContract(contractAlias) {
-    return !!this.packageData.contracts[contractAlias];
+    return this.networkFile.hasSameBytecode(contractAlias, contractClass)
```
spalladino: Shouldn't be the negation (ie contract has changed if bytecode is *not* the same)? More important: why didn't any test catch this?

src/scripts/status.js
```diff
@@ -77,50 +75,42 @@ async function contractsInfo(appController) {
...
-  const networkStdlib = appController.networkPackage.stdlib;
-  const hasNetworkStdlib = !_.isEmpty(networkStdlib);
-  const networkStdlibMatches = hasNetworkStdlib && networkStdlib.name === requiredStdlib.name && networkStdlib.version === requiredStdlib.version;
-  const hasCustomDeploy = hasNetworkStdlib && networkStdlib.customDeploy;
-  const customDeployMatches = hasCustomDeploy && networkStdlibMatches;
-
-  if (!requiresStdlib) {
-    const deployedWarn = hasNetworkStdlib ? `(though ${networkStdlib.name}@${networkStdlib.version} is currently deployed)` : '';
+  if (!packageFile.hasStdlib()) {
+    const deployedWarn = networkFile.hasStdlib() ? `(though ${networkFile.stdlibName}@${networkFile.stdlibVersion} is currently deployed)` : '';
```
spalladino: Nice :-)

spalladino:
I've just realised that many of the remaining PRs depend on this one, so any change here involves rebasing all the other PRs. Feel free to merge anyway if you think no issues are critical, and fix them in a later PR on top of the others.

src/models/files/ZosNetworkFile.js
facuspagnuolo: I'm using it in some tests, but I think I can manage with `addProxy`

src/models/files/ZosPackageFile.js
facuspagnuolo: The idea of encapsulating this behavior in a method is not to repeat it every time. If we don't have this method we will be repeating the comparison `.name === x` in many places. Of course, we could work with JS plain objects, but I think if we do this way we will reduce a lot the amount of duplicated code and it will let us refactor/modify that code in a faster way

src/models/network/NetworkAppController.js
facuspagnuolo: No need given we are returning an async call

src/scripts/status.js
facuspagnuolo: Maybe in this kind of scenarios, it is easier to see how encapsulating certain functionality under an object, instead of using a plain JS one, is more helpful.

src/models/network/NetworkBaseController.js
facuspagnuolo: Oh! `hasSameBytecode` was actually checking if both bytecodes were different 😅 
I'll fix that

src/models/files/ZosNetworkFile.js
facuspagnuolo: Yes, I noticed that after pushing this PR and fixed it in a following one (see https://github.com/zeppelinos/zos-cli/pull/261/files#diff-a2f915840192bab0d44e9d30d58495d9R93)

facuspagnuolo:
I'm closing this PR and pushing the last commit to the following one (https://github.com/zeppelinos/zos-cli/pull/258)

PR #258, opened 2018-06-17
Improve script functions using new package/network file models

facuspagnuolo:
Depends on https://github.com/zeppelinos/zos-cli/pull/257

### DIFF
--- a/test/setup.js
+++ b/test/setup.js
@@ -1,12 +1,21 @@
 import Stdlib from '../src/models/stdlib/Stdlib'
+import ZosPackageFile from '../src/models/files/ZosPackageFile'
+import ZosNetworkFile from '../src/models/files/ZosNetworkFile'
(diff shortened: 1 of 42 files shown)


test/setup.js
```diff
@@ -1,13 +1,22 @@
...
 
 require('chai')
   .use(require('chai-as-promised'))
   .use(require('zos-lib').assertions)
   .should()
 
+function useTestZosPackageFile() {
+  ZosPackageFile.prototype.write = () => {}
+  ZosNetworkFile.prototype.write = () => {}
+  ZosPackageFile.prototype.ifMissingThrow = () => {}
```
spalladino: Why is this throw removed? I'd rather modify the classed being tested as little as possible while testing.

facuspagnuolo: Agree. I'm removing this line and forcing the init script instead when needed

PR #259, opened 2018-06-18
Print help if 'zos' is invoked with no additional arguments.

### LINKED ISSUE #235: Print a brief command summary if ran with no arguments
Currently if you just type "zos" with no other command line arguments it just exist back to the command line prompt.  It would be helpful to have a brief command summary like other Linux command line utilities have.

pinkiebell:
Fixes #235

### DIFF
--- a/src/bin/errors.js
+++ b/src/bin/errors.js
@@ -8,9 +8,4 @@ module.exports = function registerErrorHandler(program) {
     console.error(`Invalid command: ${program.args.join(' ')}\nSee --help for a list of available commands.`)
     process.exit(1)
   })
-
-  program.parse(process.argv)
-  if (program.args.length === 0) {
-    program.emit('--help')
-  }
 }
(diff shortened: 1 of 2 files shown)


facuspagnuolo (approved):
Thanks @pinkiebell

src/bin/errors.js
```diff
@@ -8,4 +8,9 @@ module.exports = function registerErrorHandler(program) {
     console.error(`Invalid command: ${program.args.join(' ')}\nSee --help for a list of available commands.`)
     process.exit(1)
   })
+
+  program.parse(process.argv)
```
facuspagnuolo: We shouldn't be parsing args here, we are already doing that in `bin/zos-cli.js`. Thus we should eval if no args were given there.

src/bin/errors.js
```diff
@@ -8,4 +8,9 @@ module.exports = function registerErrorHandler(program) {
...
+  if (program.args.length === 0) {
+    program.emit('--help')
```
facuspagnuolo: you can use `program.help()` directly

facuspagnuolo (changes requested):
Thanks @pinkiebell! Left some comments

facuspagnuolo (approved):

Issue #206, opened 2018-05-22
Consider writing the proxy deployment address in the build json file


spalladino:
When deploying a contract instance, Truffle writes a `deployedAddress` (disclaimer: actual field name may be different) field in the JSON file of a contract in the build folder. This enables retrieving an instance via `MyContract.deployed()`. 

Evaluate writing such field upon create-proxy, to allow users to work with the `deployed()` method. Check how truffle handles deployment on different networks here.

makoto:
This feature is probably important to support framework like https://github.com/trufflesuite/drizzle where it wraps web3 object under redux store so I cannot instantiate contract with ZOS proxy id.

roschler:
Yes, desperately need this!  I'm currently trying crazy stuff like this in an attempt to be able debug Proxy contracts from a Truffle console session because I can't use the `deployed()` interface:

https://github.com/trufflesuite/truffle/issues/988

roschler:
@spalladino Is there anything I can do myself in the meantime to fill in that deployed address, using the addresses in the `zos.<network>.json` file, something I can edit manually?  I don't care if I have to do it every time I deploy.

roschler:
@facuspagnuolo Thanks.

PR #267, opened 2018-06-20
Store proxy address in contract build files

facuspagnuolo:
Fixes https://github.com/zeppelinos/zos-cli/issues/206

### DIFF
--- a/test/scripts/create.test.js
+++ b/test/scripts/create.test.js
@@ -2,7 +2,7 @@
 require('../setup')
 
 import CaptureLogs from '../helpers/captureLogs';
-import { Contracts, Logger } from 'zos-lib';
+import { Contracts, Logger, FileSystem as fs } from 'zos-lib';
 
 import add from '../../src/scripts/add.js';
 import push from '../../src/scripts/push.js';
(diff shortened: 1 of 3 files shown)


test/scripts/create.test.js
```diff
@@ -66,6 +66,16 @@ contract('create script', function([_, owner]) {
...
+  it('should record the proxy deployed address in contract build json file', async function () {
+    await createProxy({ contractAlias, packageFileName, network, txParams });
+
+    const contractData = fs.parseJson(`${process.cwd()}/build/contracts/${contractName}.json`)
+    const proxyData = fs.parseJson(networkFileName)
+    const proxyAddress = proxyData.proxies[contractAlias][0].implementation
+    const network_id = Contracts.getFromLocal(contractName).network_id
+    contractData.networks[network_id].address.should.be.eq(proxyAddress)
+  });
+
```
spalladino: I'd suggest adding a test to check that `Contract.deployed()` resolves correctly, since it's the actual feature we want to enable

roschler:
Is this working now so we can use the "deployed()" interface in a Truffle console session.  I clicked on @spalladino 's commit and GitHub said it couldn't find it.

Issue #268, opened 2018-06-24
Allow for changing of build directory


thecalvinchan:
truffle allows us to set a `contracts_build_directory` parameter that changes where compiled contracts are outputted. when running `zos add`, this parameter is respected in the compilation phase (because we are using truffle to compile contracts), but then fails to find the compiled contract when attempting to add the contract

https://github.com/zeppelinos/zos-cli/blob/development/src/models/local/LocalBaseController.js#L43

should be changed to respect a `contracts_build_directory` as well

PR #271, opened 2018-06-25
Added method to check for contracts_build_directory in truffle.js

thecalvinchan:
Addresses https://github.com/zeppelinos/zos-cli/issues/268

Truffle lets you set custom build directory: https://truffleframework.com/docs/advanced/configuration#contracts_build_directory

Currently, zos assumes `./build/contracts`

### DIFF
--- a/src/models/network/NetworkBaseController.js
+++ b/src/models/network/NetworkBaseController.js
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import { bytecodeDigest } from '../../utils/digest';
+import contractBuildDirectory from '../../utils/contractBuildDirectory'
 import { Contracts, Logger, FileSystem as fs, App } from 'zos-lib';
 
 const log = new Logger('NetworkController');
@@ -83,7 +84,7 @@ export default class NetworkBaseController {
   }
 
   async uploadContract(contractAlias, contractName) {
-    const contractClass = Contracts.getFromLocal(contractName);
+    const contractClass = Contracts._getFromBuildDir(contractBuildDirectory(), contractName);
     log.info(`Uploading ${contractName} contract as ${contractAlias}`);
     const contractInstance = await this.setImplementation(contractClass, contractAlias);
     this.networkPackage.contracts[contractAlias] = {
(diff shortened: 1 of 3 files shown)


src/models/network/NetworkBaseController.js
```diff
@@ -83,7 +84,8 @@ export default class NetworkBaseController {
   }
 
   async uploadContract(contractAlias, contractName) {
-    const contractClass = Contracts.getFromLocal(contractName);
+    console.log(contractBuildDirectory())
+    const contractClass = Contracts._getFromBuildDir(contractBuildDirectory(), contractName);
```
thecalvinchan: i understand we shouldn't be calling `_getFromBuildDir` as that's meant to be a private function, but perhaps this can lead to a proposed change in `zos-lib` as well to make that `getFromBuildDir` public function

src/models/network/NetworkBaseController.js
```diff
@@ -83,7 +84,7 @@ export default class NetworkBaseController {
   }
 
   async uploadContract(contractAlias, contractName) {
-    const contractClass = Contracts.getFromLocal(contractName);
+    const contractClass = Contracts._getFromBuildDir(contractBuildDirectory(), contractName);
```
thecalvinchan: i understand we shouldn't be calling `_getFromBuildDir` as that's meant to be a private function, but perhaps this can lead to a proposed change in `zos-lib` as well to make that `getFromBuildDir` public function

thecalvinchan:
Strange, running a test on the specific test that fails

`yarn test test/models/Truffle.test.js` does not fail, but running the entire suite does.

Is there some cleanup logic somewhere I should be looking at?

src/models/network/NetworkBaseController.js
thecalvinchan: Actually, if we can get this merged in first, then that would solve this issue as well:

https://github.com/zeppelinos/zos-lib/pull/183

thecalvinchan:
https://github.com/zeppelinos/zos-lib/pull/183

Could be useful to merge this in to resolve the issue of calling `_getFromBildDir`

facuspagnuolo:
Once `zos-lib` is released, we should use `Contracts#getLocalPath` in `zos-cli`, then this PR shouldn't be necessary.

facuspagnuolo:
Additionally, I think you missed adding `truffle-config` as a dependency. Please do not forget to make sure this works the way you're expecting 🙏

facuspagnuolo:
Closing this one, based on https://github.com/zeppelinos/zos-lib/pull/202 we will need just to update the dependency. We will do as soon as we release a new minor version for `zos-lib`

PR #276, opened 2018-06-26
Polish session command

### LINKED ISSUE #122: Implement a `session` command

facuspagnuolo:
Fixes #122

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules
 zos*
+.zos.session
 build
 lib
 .node-xmlhttprequest-sync-*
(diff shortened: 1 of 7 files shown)


facuspagnuolo:
@spalladino would you mind taking another look here? I added a default timeout, in a future we can add a new argument to the session command to use a custom timeout. But I'd prefer analysing the set of commands we have now before moving forward

spalladino:
LGTM. Should we add a notice whenever the user runs a command if a session is in context? Something like "Running in session network foo".

PR #281, opened 2018-06-29
Add transactions timeout to all commands

spalladino:
Add new --timeout flag to all commands that interact with the
blockchain, which is used for timing out on transactions (not
deployments).

Fixes https://github.com/zeppelinos/zos-lib/issues/149

### DIFF
--- a/src/utils/runWithTruffle.js
+++ b/src/utils/runWithTruffle.js
@@ -1,12 +1,17 @@
 import Truffle from '../models/truffle/Truffle';
 import Session from '../models/network/Session'
+import Contracts from 'zos-lib/lib/utils/Contracts';
+import _ from 'lodash';
 
-export default async function runWithTruffle(script, network, compile = false) {
+const DEFAULT_TIMEOUT = 10 * 60; // 10 minutes
+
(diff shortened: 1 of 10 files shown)


src/utils/runWithTruffle.js
```diff
@@ -1,12 +1,17 @@
 import Truffle from '../models/truffle/Truffle';
 import Session from '../models/network/Session'
+import Contracts from 'zos-lib/lib/utils/Contracts';
+import _ from 'lodash';
 
-export default async function runWithTruffle(script, network, compile = false) {
+const DEFAULT_TIMEOUT = 10 * 60; // 10 minutes
+
+export default async function runWithTruffle(script, network, { compile = false, timeout = null }) {
```
facuspagnuolo: why not using `DEFAULT_TIMEOUT` here instead?

spalladino: I was unsure if `commander` would consistently return `undefined` (and not `null`), thus causing the default value not to be used.

facuspagnuolo: I gave it a try and I guess I got an `undefined` but let's be sure about it

PR #282, opened 2018-06-29
Allow stdlib versions to be specified with npm range syntax

### LINKED ISSUE #217: Improve stdlib version validation
Currently, we are raising an error if the requested stdlib to be linked does not match exactly the package.json version of the installed NPM dependency, or if the zos.json version doesn't match it, or if the zos.<network>.json doesn't match it.

We are considering supporting just major and minor versions, but allowing to differ on patch versions.

spalladino:
An stdlib version can now be set using npm semver range syntax,
such as ^1.0.0 or ~1.3.0. When linking to a stdlib, the zos.json
file will store the range requirement, and the zos.network.json
file will act as a sort of "lock", storing the actual version
used for deployment. If the actual version does not match the
requirement, an error is thrown.

Fixes #217

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "commander": "^2.15.1",
     "lodash": "^4.17.5",
     "npm-programmatic": "0.0.10",
+    "semver": "^5.5.0",
     "truffle": "^4.1.6",
     "truffle-config": "^1.0.4",
     "truffle-core": "^4.1.8",
--- a/src/models/network/NetworkAppController.js
+++ b/src/models/network/NetworkAppController.js
@@ -163,14 +163,20 @@ export default class NetworkAppController extends NetworkBaseController {
 
     const stdlibName = this.packageFile.stdlibName;
     log.info(`Connecting to public deployment of ${stdlibName} in ${this.networkFile.network}`);
-    const stdlibAddress = Stdlib.fetch(stdlibName, this.packageFile.stdlibVersion, this.networkFile.network);
+
+    const { address: stdlibAddress, version: stdlibVersion } = Stdlib.fetch(stdlibName, this.packageFile.stdlibVersion, this.networkFile.network);
     const currentStdlibAddress = await this.app.currentStdlib()
 
-    if(stdlibAddress !== currentStdlibAddress) {
+    if (stdlibAddress !== currentStdlibAddress) {
       await this.app.setStdlib(stdlibAddress);
-      this.networkFile.stdlib = { address: stdlibAddress, ... this.packageFile.stdlib };
+      this.networkFile.stdlib = { 
+         ... this.packageFile.stdlib,
+         address: stdlibAddress, 
+         version: stdlibVersion
+      };
+    } else {
+      log.info(`Current application is already linked to stdlib ${stdlibName} at ${stdlibAddress} in ${this.network}`);
     }
-    else log.info(`Current application is already linked to stdlib ${stdlibName} at ${stdlibAddress} in ${this.network}`);
   }
 
   isStdlibContract(contractAlias) {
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -34,6 +34,7 @@
     "commander": "^2.15.1",
     "lodash": "^4.17.5",
     "npm-programmatic": "0.0.10",
+    "semver": "^5.5.0",
```
facuspagnuolo: There were no updates on the `package-lock`? It seems we were already depending on this one

src/models/stdlib/Stdlib.js
```diff
@@ -45,9 +50,12 @@ export default class Stdlib {
...
-    const packageVersion = this.getPackage().version;
-    this.version = version || packageVersion
+    this.version = version
     this.nameAndVersion = nameAndVersion
-    if (this.version !== packageVersion) throw Error(`Requested stdlib version ${version} does not match stdlib package version ${packageVersion}`)
+
+    const packageVersion = this.getPackage().version
+    if (!Stdlib.satisfies(packageVersion, version)) {
+      throw Error(`Requested stdlib version ${version} does not match stdlib package version ${packageVersion}`)
+    }
```
facuspagnuolo: Should we add a validation method besides `satisfies`? This is actually the logic we are needing here and in `StdlibProvider`

src/models/network/NetworkAppController.js
```diff
@@ -171,13 +171,18 @@ export default class NetworkAppController extends NetworkBaseController {
...
+    const { address: stdlibAddress, version: stdlibVersion } = Stdlib.fetch(stdlibName, this.packageFile.stdlibVersion, this.networkFile.network);
 
-    if(stdlibAddress !== currentStdlibAddress) {
+    if (stdlibAddress !== currentStdlibAddress) {
       await this.app.setStdlib(stdlibAddress);
-      this.networkFile.stdlib = { address: stdlibAddress, ... this.packageFile.stdlib };
+      this.networkFile.stdlib = { 
+         ... this.packageFile.stdlib,
+         address: stdlibAddress, 
+         version: stdlibVersion
```
facuspagnuolo: which attributes are we writing from `packageFile.stdlib`? I think it is just the `name`, right? In case I'm not wrong, I think it would be clearer making it explicit, do you agree?

test/scripts/push.test.js
```diff
@@ -167,6 +167,22 @@ contract('push script', function([_, owner]) {
     });
   };
 
+  const shouldSetStdlib = function () {
+    const stdlibAddress = '0x0000000000000000000000000000000000000010';
+
+    it('should set stdlib in deployed app', async function () {
+      const app = await App.fetch(this.networkFile.appAddress);
+      const stdlib = await app.currentStdlib();
```
facuspagnuolo: no need to await here

test/models/Stdlib.test.js
```diff
@@ -50,8 +52,8 @@ contract('Stdlib', function () {
       this.stdlib.name.should.eq(packageName)
     })
 
-    it('should set a version', async function () {
-      this.stdlib.version.should.eq('1.1.0')
+    it('should not set a version if not specified', async function () {
+      should.not.exist(this.stdlib.version)
```
facuspagnuolo: I'm not sure about this scenario. It's kind of weird installing a dependency and not seeing which version I got installed in my package-file. Maybe because I'm used to those package managers that assume you requested the latest version if you didn't specify one. For example, NPM will assume you're requesting the latest including minor and patch updates. Do you agree?

src/models/network/NetworkAppController.js
spalladino: It's also the `customDeployment` flag. I'd rather ensure all info is copied.

spalladino:
Implemented changes and rebased on master. @facuspagnuolo please re-review!

test/models/Stdlib.test.js
spalladino: Sounds good!

PR #286, opened 2018-07-04
Ensure no more than one instance of CLI is running simultaneously

### LINKED ISSUE #168: Prevent two instances of CLI to be run simultaneously
CLI is not prepared to run multiple instances at the same time. Consider creating a `.zos.lock` file to signal that there is an ongoing execution, clean it up upon exit (make sure the file is deleted if the process is killed via a non-SIGKILL signal), and validate it does not exist when the CLI runs.

spalladino:
Use npm/lockfile to acquire unique lock on a .zos.lock local file and exit the CLI if cannot. Fixes #168.

### DIFF
--- a/src/bin/zos-cli.js
+++ b/src/bin/zos-cli.js
@@ -1,7 +1,17 @@
 #! /usr/bin/env node
 
 import { Logger } from 'zos-lib'
+import { lockSync } from 'lockfile';
 import program from './program'
+
+// Acquire file lock to ensure no other instance is running
+const LOCKFILE = '.zos.lock';
+try { 
+  lockSync(LOCKFILE, { retries: 0 }) 
+} catch (e) { 
+  console.error(`Cannot run more than one instance of 'zos' at the same time.\nIf you are sure that no other instances are actually running, manually remove the file ${LOCKFILE} and try again.`);
+  process.exit(1); 
+}
 
 Logger.silent(false)
 program.parse(process.argv)
--- a/package-lock.json
+++ b/package-lock.json
@@ -5430,6 +5430,14 @@
         }
       }
     },
+    "lockfile": {
+      "version": "1.0.4",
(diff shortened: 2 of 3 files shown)


src/bin/zos-cli.js
```diff
@@ -1,8 +1,18 @@
...
 import { Logger } from 'zos-lib'
+import { lockSync } from 'lockfile';
 import program from './program'
 
+// Acquire file lock to ensure no other instance is running
+const LOCKFILE = '.zos.lock';
+try { 
+  lockSync(LOCKFILE, { retries: 0 }) 
+} catch (e) { 
+  console.error(`Could lock file '${LOCKFILE}'. Ensure no other instance of zos is running.`);
```
facuspagnuolo: I'd prefer something clearer like `Cannot run more than one instance of zos at the same time. If you are sure you are not running more than once, please remove .zos.lock` or sth like that

src/bin/zos-cli.js
```diff
@@ -1,8 +1,18 @@
...
 import program from './program'
 
+// Acquire file lock to ensure no other instance is running
+const LOCKFILE = '.zos.lock';
+try { 
+  lockSync(LOCKFILE, { retries: 0 }) 
+} catch (e) { 
+  console.error(`Could lock file '${LOCKFILE}'. Ensure no other instance of zos is running.`); 
+  process.exit(1); 
+}
```
facuspagnuolo: Don't we need to clean the file?

spalladino: Nope, `lockfile` takes care of everything :-)

src/bin/zos-cli.js
spalladino: Updated to `Cannot run more than one instance of 'zos' at the same time.\nIf you are sure that no other instances are actually running, manually remove the file ${LOCKFILE} and try again.`

spalladino:
Changed the error message, but I'd keep the lockfile package as mentioned above.

PR #299, opened 2018-07-05
Fix proxies deployed integration with truffle

facuspagnuolo:


facuspagnuolo:
I tried but again Truffle is headache my friend :(

PR #310, opened 2018-07-08
Fix compare command

### LINKED ISSUE #307: Compare and pull commands do not automatically exit
It requires an interruption to actually exit the process.

### LINKED ISSUE #308: Remove timeout option from compare command
Compare does not issue any transactions, only queries. Remove the timeout option.

### LINKED ISSUE #309: Rename "expected vs actual" to "on-chain vs local" for compare command
The compare command shows all the findings comparing what was expected vs what it found actually. We should rename those entries to what was found on-chain and what was found locally instead.

s1na:
- Removes `--timeout` option (Fixes #308)
- Renames expected and actual to on-chain and local (Fixes #309)
- Fixes `EventsFilter` to clear timeout when the main promise resolves, which caused compare and pull to hang (Fixes #307)

### DIFF
--- a/src/models/status/StatusReport.js
+++ b/src/models/status/StatusReport.js
@@ -7,7 +7,7 @@ export default class StatusReport {
 
   log(logger) {
     logger.error(this.description)
-    logger.error(` - on-chain: ${this.expected}`)
-    logger.warn(` - local:   ${this.observed}\n`)
+    logger.error(` - local: ${this.expected}`)
+    logger.warn(` - on-chain:   ${this.observed}\n`)
   }
 }
--- a/src/models/status/StatusReport.js
+++ b/src/models/status/StatusReport.js
@@ -7,7 +7,7 @@ export default class StatusReport {
 
   log(logger) {
     logger.error(this.description)
-    logger.error(` - expected: ${this.expected}`)
-    logger.warn(` - actual:   ${this.observed}\n`)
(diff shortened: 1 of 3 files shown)


src/models/status/StatusReport.js
```diff
@@ -7,7 +7,7 @@ export default class StatusReport {
 
   log(logger) {
     logger.error(this.description)
-    logger.error(` - expected: ${this.expected}`)
-    logger.warn(` - actual:   ${this.observed}\n`)
+    logger.error(` - on-chain: ${this.expected}`)
+    logger.warn(` - local:   ${this.observed}\n`)
```
facuspagnuolo: This not totally right due to a bad naming of the `StatusReport` variables. Note that we are currently using `observed` to refer to what was found on-chain and `expected` to what was found locally. Then we have two alternatives:

1- Swapping the declaration of these variables in the constructor, it should be: 

```js
export default class StatusReport {
  constructor(observed, expected, description) {
    this.observed = observed
    this.expected = expected
    this.description = description
  }
...
}
```

But you will have to fix many tests. I understand if you don't want to do this.

2- We can fix your change to use `observed` to output whatever was found **on-chain** and `expected` to show what was found **locally**. And we can fix the naming issue later.

facuspagnuolo:
Thanks a lot for the contribution @s1na !

PR #318, opened 2018-07-14
Support creating non-upgradeable contract instances

facuspagnuolo:
Fixes https://github.com/zeppelinos/zos-lib/issues/32

Analyzing this issue a bit, I thought about two alternatives:

### Using on-chain version of a contract 
The idea of this approach is to allow users to create instances of any contract they have already registered in their `App`s. 

**PRO:** We will be using exactly the same on-chain bytecode
**CON:** Currently, every non-upgradeable instance will follow the current initializers model. This means, even if a user could create an instance of a contract using a constructor function, given that we are registering only contracts without constructors, we will be forcing them to work with initializers.

### Using local version of a contract
This approach could allow users to create instances of any contract they have locally. 

**PRO:** It will allow users to work with constructors instead of the current initializer model.
**CON:** We will not be working with registered implementations. Additionally, we will have to modify the interface of the `create` command of the CLI to support receiving arguments without the initialize method. The last one could be cumbersome.

### Conclusion
I'd go with the first alternative. IMO, working with registered implementations is a must. Although it is not a good idea to force users to work with initializers when creating non-upgradeable instances, we are working on a new initializers model that will solve this issue (constructors are going to be on-chain). Having that said, let's think if we are ok with merging this PR while we keep working on the new initializers model, or if we should do it afterward.

### DIFF
--- a/src/models/files/ZosNetworkFile.js
+++ b/src/models/files/ZosNetworkFile.js
@@ -1,7 +1,7 @@
 import _ from 'lodash'
+import Stdlib from '../stdlib/Stdlib'
 import { Logger, FileSystem as fs } from 'zos-lib'
 import { bytecodeDigest, bodyCode, constructorCode } from '../../utils/contracts'
-import Stdlib from '../stdlib/Stdlib';
 
 const log = new Logger('ZosNetworkFile')
 
@@ -12,7 +12,7 @@ export default class ZosNetworkFile {
     this.network = network
     this.fileName = fileName
 
-    const defaults = this.packageFile.isLib ? { contracts: {}, lib: true, frozen: false } : { contracts: {}, proxies: {}, nonUpgradeableInstances: {} }
+    const defaults = this.packageFile.isLib ? { contracts: {}, lib: true, frozen: false } : { contracts: {}, instances: {} }
     this.data = fs.parseJsonIfExists(this.fileName) || defaults
   }
 
@@ -65,11 +65,15 @@ export default class ZosNetworkFile {
   }
 
   get proxies() {
-    return this.data.proxies || {}
+    return this.instanceAliases.reduce((previous, alias) => {
+      const instances = this.instancesOf(alias)
+      if (instances.length >= 0) previous[alias] = instances
+      return previous
+    }, {})
   }
 
-  get nonUpgradeableInstances() {
-    return this.data.nonUpgradeableInstances || {}
+  get instances() {
+    return this.data.instances || {}
   }
 
   get contracts() {
@@ -80,8 +84,8 @@ export default class ZosNetworkFile {
     return Object.keys(this.contracts)
   }
 
-  get proxyAliases() {
-    return Object.keys(this.proxies)
+  get instanceAliases() {
+    return Object.keys(this.instances)
   }
 
   get isLib() {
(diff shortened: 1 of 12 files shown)


src/models/files/ZosNetworkFile.js
```diff
@@ -67,6 +67,10 @@ export default class ZosNetworkFile {
     return this.data.proxies || {}
   }
 
+  get nonUpgradeableInstances() {
+    return this.data.nonUpgradeableInstances || {}
```
spalladino: Just throwing it out there: how about renaming `proxies` to `instances` (yes, it is a breaking change), and adding an `upgradeable` flag on each of them? That would also help by removing the mention of proxies from the json file.

src/models/network/NetworkAppController.js
```diff
@@ -52,23 +93,26 @@ export default class NetworkAppController extends NetworkBaseController {
...
     await this.fetch();
     const contractClass = this.localController.getContractClass(contractAlias);
-    this.checkInitialization(contractClass, initMethod, initArgs);
-    const proxyInstance = await this.app.createProxy(contractClass, contractAlias, initMethod, initArgs);
-    const implementationAddress = await this.app.getImplementation(contractAlias);
-    this._updateTruffleDeployedInformation(contractAlias, proxyInstance)
+    this.checkInitialization(contractClass, initMethod);
+    const instance = upgradeable
+      ? await this.app.createProxy(contractClass, contractAlias, initMethod, initArgs)
+      : await this.app.createNonUpgradeableInstance(contractClass, contractAlias, initMethod, initArgs)
```
spalladino: I'd consider pushing this ternary operation down to zos-lib `App`, so this becomes:
```js
const proxyInstance = await this.app.createInstance(upgradeable, contractClass, contractAlias, initMethod, initArgs);
```

facuspagnuolo: I like the idea, in the meantime, I pushed this PR in order to have this logic in `zos-lib`:  https://github.com/zeppelinos/zos-lib/pull/209

src/models/files/ZosNetworkFile.js
facuspagnuolo: I started doing exactly what you suggest, but it was a heavy braking change so I decided to stop to discuss it with you guys before moving forward.

src/models/network/NetworkAppController.js
```diff
@@ -1,10 +1,51 @@
...
+  // push1 00     | 0x60 0x00  | 0xCS 0x00
+  // return       | 0xF3       |
+
+  const ASM_CODE_COPY = `0x73${implementationAddress}803b8091600080913c6000f3`;
+
+  const params = Object.assign({}, contractClass.defaults(), this.txParams, { to: 0x0, data: ASM_CODE_COPY })
+  const txHash = await web3.eth.sendTransaction(params)
+  const receipt = await web3.eth.getTransactionReceipt(txHash)
+  const instance = contractClass.at(receipt.contractAddress)
+  // log.info(`${contractName} instance created at ${instance.address}`)
```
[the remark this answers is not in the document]
facuspagnuolo: I missed those, thanks!

src/models/network/NetworkAppController.js
```diff
@@ -1,10 +1,51 @@
...
+
+  if(typeof(initArgs) !== 'undefined') {
+    // this could be front-run
```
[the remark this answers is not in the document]
facuspagnuolo: Sure! Please check https://github.com/zeppelinos/zos-lib/pull/209

src/models/files/ZosNetworkFile.js
```diff
@@ -243,6 +259,15 @@ export default class ZosNetworkFile {
     if(this.proxiesOf(alias).length === 0) delete this.data.proxies[alias]
   }
 
+  setNonUpgradeableInstances(alias, value) {
```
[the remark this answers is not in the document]
facuspagnuolo: Thanks!

facuspagnuolo:
@spalladino left another commit unifying non-upgradeable instances and proxies in the network file. Please take another look!

src/models/files/ZosNetworkFile.js
```diff
@@ -64,11 +64,15 @@ export default class ZosNetworkFile {
   }
 
   get proxies() {
-    return this.data.proxies || {}
+    return this.instanceAliases.reduce((previous, alias) => {
+      const instances = this.instancesOf(alias)
```
spalladino: Shouldn't be `proxiesOf` instead of `instancesOf`?

spalladino: Also, this entire function should be implementable with a lodash one-liner, something like:
```js
_.mapValues(this.instanceAliases, (instances) => _.filter(instances, 'upgradeable'))
```

spalladino:
Looks good, spotted one issue only. I'd also consider adding tests for upgrade and upgrade-all, to ensure it does not attempt to pick up nonUpgradeableInstances for upgrading.

spalladino:
Also: should we merge this into `develop`, as it is a breaking change?

facuspagnuolo:
@spalladino agree, let's merge it against develop

facuspagnuolo:
@spalladino would you mind taking another look?

PR #322, opened 2018-07-17
Fail with explicit error when creating proxy for missing stdlib

### LINKED ISSUE #218: More explicit failure when a proxy is being created for an implementation that doesn't exist
If you try to create a proxy for a PackagedApp that is linked to a stdlib incorrectly (i.e. linked but not pushed), createProxyAndCall will revert, not because of problems with initialization but with the creation of the proxy itself (because the implementation is 0x0).

It would be super handy for DX to have the CLI state that the proxy cannot be created because there is no known implementation for that alias. This check should be done outside of the json files.

spalladino:
Creating a proxy for a contract that belongs to an stdlib that was
linked, but that link was not pushed to the network, would yield
a "VM Revert" error. Now it explains that the stdlib is linked but
a push is missing.

Fixes #218

### DIFF
--- a/src/models/network/NetworkBaseController.js
+++ b/src/models/network/NetworkBaseController.js
@@ -111,40 +111,49 @@ export default class NetworkBaseController {
   }
 
   checkLocalContractsDeployed(throwIfFail = false) {
-    let msg;
+    this._handleErrorMessage(this._errorForLocalContractsDeployed(), throwIfFail);
+  }
+
+  _errorForLocalContractsDeployed() {
     const [contractsDeployed, contractsMissing] = _.partition(this.packageFile.contractAliases, (alias) => this.isContractDeployed(alias));
     const contractsChanged = _.filter(contractsDeployed, (alias) => this.hasContractChanged(alias));
 
     if (!_.isEmpty(contractsMissing)) {
-      msg = `Contracts ${contractsMissing.join(', ')} are not deployed.`;
+      return `Contracts ${contractsMissing.join(', ')} are not deployed.`;
     } else if (!_.isEmpty(contractsChanged)) {
-      msg = `Contracts ${contractsChanged.join(', ')} have changed since the last deploy.`;
+      return `Contracts ${contractsChanged.join(', ')} have changed since the last deploy.`;
     }
-
-    if (msg && throwIfFail) throw Error(msg);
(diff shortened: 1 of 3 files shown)


src/models/network/NetworkBaseController.js
```diff
@@ -111,48 +111,57 @@ export default class NetworkBaseController {
...
+    return this.packageFile.hasContract(contractAlias)
+  }
+
   isContractDefined(contractAlias) {
     return this.packageFile.hasContract(contractAlias)
   }
 
   isContractDeployed(contractAlias) {
-    return !this.packageFile.hasContract(contractAlias) || this.networkFile.hasContract(contractAlias);
+    return !this.isLocalContract(contractAlias) || this.networkFile.hasContract(contractAlias);
```
facuspagnuolo: I'd move all these logic somewhere else, half of the lines of this Controller are related to contract deployment verifications

src/models/network/NetworkBaseController.js
```diff
@@ -111,48 +111,57 @@ export default class NetworkBaseController {
   }
 
   checkLocalContractsDeployed(throwIfFail = false) {
-    let msg;
+    this._handleErrorMessage(this._errorForLocalContractsDeployed(), throwIfFail);
```
facuspagnuolo: IMO checking and handling are kind of disruptive here, it is kind of weird to call a handler as a first step. I would have expected a condition at least.

spalladino: Agree, changed it

src/models/network/NetworkBaseController.js
spalladino: Not sure if I agree. All verifications require access to both files, and are invoked by `script`s which typically interact with the network controller directly. Furthermore, logic is different between app and lib, which matches nicely with the hierarchy of base/app/lib controllers.

I can give it a shot by moving this to a ContractChecker or DeploymentVerifier, but I doubt it will make the design more clear, as the Controller will need to pass-through most calls to it.

spalladino:
@facuspagnuolo implemented one of the changes, though I'm not sure [about the other](https://github.com/zeppelinos/zos-cli/pull/322#discussion_r204433888). I can give it a shot, but I'm not sure it's worthwhile atm. WDYT?

PR #324, opened 2018-07-18
Refactor command options

### LINKED ISSUE #302: Improve helper details for timeout flag
We should make explicit the timeout is for every single transaction being processed when a command is run. Additionally, we should show the default value we are considering.

spalladino:
Refactor common options into helpers. Also fixes #302.

### DIFF
--- a/src/bin/options.js
+++ b/src/bin/options.js
@@ -1,0 +1,15 @@
+import program from 'commander'
+
+program.Command.prototype.withNetworkOptions = function () {
+  return this
+    .option('-n, --network <network>', 'network to be used')  
+    .option('-f, --from <from>', 'specify transaction sender address')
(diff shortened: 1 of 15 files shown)


src/bin/options.js
```diff
@@ -0,0 +1,21 @@
...
+    .option('-f, --from <from>', 'specify transaction sender address')
+    .withNetworkTimeoutOption()
+};
+
+program.Command.prototype.withOptionalPushOptions = function () {
+  return this
+    .option('--push [network]', 'push all changes to the specified network')
+    .option('-f, --from <from>', 'specify the transaction sender address for --push')
+    .withNetworkTimeoutOption()
+};
```
facuspagnuolo: I'd remove the optional part from the method name here since all these options are optional

PR #327, opened 2018-07-18
Polish update command files

facuspagnuolo:
The update command related files were wrongly named. This PR renames these files (command, scripts and test files). I'm also using update instead of upgrade in many outputted messages

### DIFF
--- a/src/commands/update.js
+++ b/src/commands/update.js
@@ -1,12 +1,12 @@
 'use strict';
 
-import upgrade from '../scripts/upgrade'
+import update from '../scripts/update'
 import runWithTruffle from '../utils/runWithTruffle'
 import { parseInit } from '../utils/input'
 
 const name = 'update'
 const signature = `${name} [alias] [address]`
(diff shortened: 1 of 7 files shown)


src/commands/update.js
```diff
@@ -1,20 +1,20 @@
...
 
-import upgrade from '../scripts/upgrade'
+import update from '../scripts/update'
 import runWithTruffle from '../utils/runWithTruffle'
 import { parseInit } from '../utils/input'
 
 const name = 'update'
 const signature = `${name} [alias] [address]`
-const description = 'upgrade contract to a new logic. Provide the [alias] you added your contract with, or use --all flag to upgrade all. If no [address] is provided, all instances of that contract class will be upgraded'
+const description = 'update contract to a new logic. Provide the [alias] you added your contract with, or use --all flag to update all. If no [address] is provided, all instances of that contract class will be upgraded'
```
spalladino: _all instances of that contract class will be **updated**_

facuspagnuolo:
I really do agree with your opinion, we always talk about `upgradeable` smart contracts. Let's open an issue to discuss it later

PR #328, opened 2018-07-18
Add test for error message upon undeployed stdlib

### LINKED ISSUE #296: Incorrect error message when pushing to a network where stdlib is unavailable
Given a push to a network for which the linked stdlib has no zos.network.json file, the error message is `Required stdlib version 1.0.1 does not match stdlib package version undefined`. Instead, show something along the lines of `Required stdlib foo@1.0.1 is not deployed to network NETWORK. Consider using --deploy-stdlib for a custom deployment.`

spalladino:
Adds test for #296 which was not reproducible, so this PR adds a test to ensure it works.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5787,6 +5787,10 @@
       "version": "file:test/mocks/mock-stdlib-invalid",
       "dev": true
     },
+    "mock-stdlib-undeployed": {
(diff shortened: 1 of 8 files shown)


facuspagnuolo:
If the issue was not reproducible I'd remove the PR "Fixes" part and close the issue directly.

Issue #274, opened 2018-06-25
Add command to automate verification on an Ethereum explorer


spalladino:
Add a new command (`verify`?) to use a blockchain explorer API to automatically verify a deployed set of contracts.

frangio:
Does this refer to the logic contracts or the proxies?

Last time I checked Etherscan did not have an API for this, and the verification form has a captcha.

spalladino:
Currently @jcarpanelli is working on this

PR #339, opened 2018-07-26
Verify contracts command

jcarpanelli:
Fixes https://github.com/zeppelinos/zos-cli/issues/274

This is the first approach to a possible implementation of a contract verification on  [etherchain](https://www.etherchain.org). I still have to code some integration tests for the etherchain communication workflow part.
Some validations have been implemented in order to forbid a user to try to verify a contract without first initializing, adding and pushing a contract to a network so we can retrieve the contract address from the zos._network_.json file. Consider that etherchain only allows contract verification and publish on the mainnet, so if  other network is specified an error (_Unable to retrieve the contract creation bytecode_) will be thrown.

### DIFF
--- a/src/models/local/LocalBaseController.js
+++ b/src/models/local/LocalBaseController.js
@@ -105,6 +105,16 @@ export default class LocalBaseController {
     }
   }
 
+  getContractSourcePath(contractAlias) {
+    const contractName = this.packageFile.contract(contractAlias)
+    if (contractName) {
+      const contractDataPath = Contracts.getLocalPath(contractName)
+      return fs.parseJson(contractDataPath).sourcePath
+    } else {
+      throw Error(`Cound not find ${contractAlias} in contracts directory.`)
+    }
+  }
+
   writePackage() {
     this.packageFile.write()
   }
--- a/src/models/local/LocalBaseController.js
+++ b/src/models/local/LocalBaseController.js
@@ -110,7 +110,7 @@ export default class LocalBaseController {
     if (contractName) {
       const contractDataPath = Contracts.getLocalPath(contractName)
       const { compiler, sourcePath } = fs.parseJson(contractDataPath)
-      return { compiler, sourcePath }
+      return { sourcePath, compilerVersion: compiler.version }
     } else {
       throw Error(`Cound not find ${contractAlias} in contracts directory.`)
     }
--- a/src/models/local/LocalBaseController.js
+++ b/src/models/local/LocalBaseController.js
@@ -112,7 +112,7 @@ export default class LocalBaseController {
       const { compiler, sourcePath } = fs.parseJson(contractDataPath)
       return { sourcePath, compilerVersion: compiler.version }
     } else {
-      throw Error(`Cound not find ${contractAlias} in contracts directory.`)
+      throw Error(`Could not find ${contractAlias} in contracts directory.`)
     }
   }
 
--- a/src/models/local/LocalBaseController.js
+++ b/src/models/local/LocalBaseController.js
@@ -109,7 +109,8 @@ export default class LocalBaseController {
     const contractName = this.packageFile.contract(contractAlias)
     if (contractName) {
       const contractDataPath = Contracts.getLocalPath(contractName)
-      return fs.parseJson(contractDataPath).sourcePath
+      const { compiler, sourcePath } = fs.parseJson(contractDataPath)
+      return { compiler, sourcePath }
     } else {
       throw Error(`Cound not find ${contractAlias} in contracts directory.`)
     }
--- a/src/scripts/verify.js
+++ b/src/scripts/verify.js
@@ -1,0 +1,14 @@
+import LocalAppController from  '../models/local/LocalAppController'
+import ZosPackageFile from "../models/files/ZosPackageFile"
+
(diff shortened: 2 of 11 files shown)


src/models/local/LocalBaseController.js
```diff
@@ -105,6 +105,17 @@ export default class LocalBaseController {
     }
   }
 
+  getContractSourcePath(contractAlias) {
+    const contractName = this.packageFile.contract(contractAlias)
+    if (contractName) {
+      const contractDataPath = Contracts.getLocalPath(contractName)
+      const { compiler, sourcePath } = fs.parseJson(contractDataPath)
```
spalladino: Note that the resulting `sourcePath` does not resolve any import aliases. This means that if the contract is part of a package (such as OpenZeppelin), the source path will be `openzeppelin-solidity/contracts/MyContract.sol`, instead of the correct `node_modules/openzeppelin-solidity/contracts/MyContract.sol`.

You should be able to distinguish between the two, as local contracts have an absolute path (such as `/home/spalladino/Projects/zeppelinos/cli/contracts/mocks/ImplV1.sol`) and the ones in a node module have a relative path.

I'm ok with leaving this as-is now and handling this case in a future PR.

src/scripts/verify.js
```diff
@@ -0,0 +1,7 @@
+import ControllerFor from '../models/network/ControllerFor'
+
+export default function verify(contractAlias, { network = 'mainnet', txParams = {}, networkFile = undefined, optimizer = false, optimizerRuns = 200, remote = 'etherchain' }) {
```
spalladino: Down the road, we may want to add a flag to verify all contracts (such as a `--all` option, like in the `add` command). Let's make sure that we have an issue that capture this, along with any future improvements we may want to work on.

src/models/local/LocalBaseController.js
jcarpanelli: As `truffle-flattener` is using the `truffle-resolve` package to resolve dependencies imports, it will automatically prepend the filesystem directory path (i.e., `/home/spalladino/Projects/zeppelinos/cli/node_modules/`) to the relative path that is read from the `Contract.json` file. This is also the expected behaviour when you flatten contracts that are not inside the project `contracts` folder, for example an OpenZeppelin contract.
You can check that implementation [here](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-resolver/npm.js). Please, let me know if I'm misunderstanding you or if I'm wrong.

test/scripts/verify.test.js
```diff
@@ -1,55 +1,103 @@
...
     })
   })
+
+  describe('contract verification', function () {
+    beforeEach(async function () {
+      packageFile = new ZosPackageFile('test/mocks/packages/package-with-contracts.zos.json')
+      networkFile = packageFile.networkFile(network)
+      await push({ network, networkFile, txParams })
+      const contracts = networkFile.contracts
+      networkFile.contracts = contracts
```
spalladino: Why these two lines?

test/scripts/verify.test.js
```diff
@@ -1,55 +1,103 @@
...
+contract('verify script', function () {
   const contractAlias = 'Impl'
   const network = 'test'
   const txParams = {}
-  let packageFile;
-  let networkFile;
+  let packageFile
+  let networkFile
+  let logs
+  let axiosStub
```
spalladino: Let's use `this.VARNAME` in the beforeEach blocks instead of `let VARNAME` here, for consistency. This makes sure the tests work if we later abstract them to a behaviour separate from the setup.

spalladino:
Also: could you add an entry to the changelog.md with this new command?

xinbenlv:
Kudos for building such a wonderful feature. When is it planned to be released?

xinbenlv:
Or how can I use the dev/nightly version? @jcarpanelli  @spalladino

spalladino:
Hey @xinbenlv! We are planning on a new 1.4.0 RC early next week. If not, you can checkout the CLI and build it on your end, but I wouldn't advise this for production usage, only for testing.

xinbenlv:
Thank you, let me check how to build a dev version. Thank you for answering my question. Also, the actual question we have is that:
1. when built with ZOS proxy pattern, what's the proxy contract code that we can post and verify it on `etherscan.io`?
2. does ZOS plan to support verify on etherscan.io as oppose to etherchain.org which seems not supporting ropsten and other testnet?

spalladino:
@xinbenlv 

1. You can find it [here](https://github.com/zeppelinos/zos/blob/master/packages/lib/contracts/upgradeability/UpgradeabilityProxy.sol).

2. Our original plans were working on etherscan, but they have a captcha set up that prevents us from automating the verification process.

xinbenlv:
I see, thanks Santiago. Have etherscan told people if they have plan to support command-line verification process in the future, e.g. with a ssh public key or what other ways that both allow the verification to be fully automated and mitigate the risk of abuse and security attacks?

spalladino:
We have contacted them privately, but got no response back so far

xinbenlv:
I see, thank you, I posted in public on https://www.reddit.com/r/etherscan/comments/970voh/verification_through_api_commendline/

PR #341, opened 2018-07-26
Remove full path to truffle to fix CI

spalladino:


spalladino:
Issue was with travis cache, not with the test command

Issue #298, opened 2018-07-05
Add tests for commands parsing and option handling


spalladino:
The entry point of the CLI is an instance of a [commander](https://github.com/tj/commander.js/) `program`. The different `commands` (found in the folder with the same name) are each in a different file, where they register themselves to the `program`, manipulate the options received from commander (if needed), and delegate the logic to the code in the `scripts` folder (sometime combining more than one `script`, such as `add --push`).

Even though `scripts` are thoroughly tested, we are not currently testing the code in the `commands` folder. We need to add a suite of tests that use the commander `program` to parse a string as if it were from the command line, and use test doubles to verify that the `scripts` actually get the correct params.

Something along the lines of the following (note that the syntax is invented, and will depend on what kind of test doubles are used).
```js
it('should call add with several contracts', function () {
  program.parse('zos add foo bar')
  expect(add).to.have.been.called.with(['foo', 'bar'])
});

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 180.0 DAI (180.0 USD @ $1.0/DAI) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/zeppelinos/zos-cli/298/861).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $30,665.12 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 4 semanas, 1 día a partir de ahora. Please review their questions below:


1. [on247](https://gitcoin.co/profile/on247) has started work. 	
 * Q: Write the tests for each program

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/zeppelinos/zos-cli/298/861).

gitcoinbot:
@on247 Hello from Gitcoin Core - are you still working on this issue? Please submit a WIP PR or comment back within the next 3 days or you will be removed from this ticket and it will be returned to an ‘Open’ status. Please let us know if you have questions!
* [x] warning (3 days)
* [ ] escalation to mods (6 days)

[...]

PR #346, opened 2018-07-31
Add tests for commands parsing and option handling

bakaoh:
Fix: #298

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -22,10 +22,10 @@ const register = program => program
 async function action(name, version, options) {
   const { force } = options
   if (options.lib) {
-    if (options.stdlib) throw Error('Cannot set a stdlib in a library project')
+    if (options.link) throw Error('Cannot set a stdlib in a library project')
     await initLib({ name, version, force })
   } else {
-    const { stdlib: stdlibNameVersion, install: installLib } = options
+    const { link: stdlibNameVersion, install: installLib } = options
     await init({ name, version, stdlibNameVersion, installLib, force })
   }
   await push.tryAction(options)
--- a/test/commands/add.test.js
+++ b/test/commands/add.test.js
@@ -1,22 +1,22 @@
 'use strict'
 require('../setup')
 
-import {setup, itShouldParse} from './share';
+import {stubCommands, itShouldParse} from './share';
 
 contract('add command', function() {
 
-  setup()
+  stubCommands()
 
-  itShouldParse('should call add script with an alias and a filename', 'add', ['node', 'zos', 'add', 'ImplV1:Impl', '--skip-compile'], function(fake) {
-    fake.should.have.been.calledWithExactly({ contractsData: [ { name: 'ImplV1', alias: 'Impl' } ] })
+  itShouldParse('should call add script with an alias and a filename', 'add', 'zos add ImplV1:Impl --skip-compile', function(add) {
+    add.should.have.been.calledWithExactly({ contractsData: [ { name: 'ImplV1', alias: 'Impl' } ] })
   })
   
-  itShouldParse('should call add-all script when passing --all option', 'addAll', ['node', 'zos', 'add', '--all', '--skip-compile'], function(fake) {
-    fake.should.have.been.calledWithExactly( { } )
+  itShouldParse('should call add-all script when passing --all option', 'addAll', 'zos add --all --skip-compile', function(addAll) {
+    addAll.should.have.been.calledWithExactly( { } )
   })
 
-  itShouldParse('should call push script when passing --push option', 'push', ['node', 'zos', 'add', '--all', '--push', 'test', '--skip-compile'], function(fake) {
-    fake.should.have.been.calledWithExactly({ deployStdlib: undefined, reupload: undefined, network: 'test', txParams: {} })
+  itShouldParse('should call push script when passing --push option', 'push', 'zos add --all --push test --skip-compile', function(push) {
+    push.should.have.been.calledWithExactly({ deployStdlib: undefined, reupload: undefined, network: 'test', txParams: {} })
   })
 
 })
--- a/test/commands/add.test.js
+++ b/test/commands/add.test.js
@@ -1,0 +1,64 @@
+'use strict'
+require('../setup')
+
+import sinon from 'sinon';
+
+import * as add from '../../src/scripts/add';
+import * as addAll from '../../src/scripts/add-all';
+import * as push from '../../src/scripts/push';
+import * as runWithTruffle from '../../src/utils/runWithTruffle'
+import Session from '../../src/models/network/Session'
+import program from '../../src/bin/program';
+
+contract('add command', function() {
+  
+  it('should call add script with an alias and a filename', function(done) {
+    var addFake = sinon.fake(function () {
+      expect(addFake.calledOnce).to.equal(true)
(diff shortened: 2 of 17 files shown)


spalladino:
Awesome work @bakaoh, thanks a lot! I'd refactor the code to reduce repetition and make more usage of before/after blocks, thus making the tests themselves as slim as possible.

So, instead of doing the following:
```js
it('should call add script with an alias and a filename', function(done) {
  var addFake = sinon.fake(function () {
    expect(addFake.calledOnce).to.equal(true)
    expect(addFake.calledWithExactly({ contractsData: [ { name: 'ImplV1', alias: 'Impl' } ] })).to.equal(true)
    addStub.restore()
    done()
  })
  var addStub = sinon.stub(add, 'default').callsFake(addFake)
  program.parse(['node', 'zos', 'add', 'ImplV1:Impl', '--skip-compile'])
});
```

I'd try to build something like this, where the before and after blocks can be reused by other tests:
```js
beforeEach('set up stubs', function () {
  this.addFake = sinon.fake();
  this.addStub = sinon.stub(add, 'default').callsFake(addFake);
});

afterEach('restore', function () {
  this.addStub.restore();
});

it('should call add script with an alias and a filename', function(done) {
  program.parse(['node', 'zos', 'add', 'ImplV1:Impl', '--skip-compile'])
  this.addStub.should.have.been.calledOnceWithExactly({ contractsData: [ { name: 'ImplV1', alias: 'Impl' } ] })
});
```

(I'm using [sinon-chai](https://github.com/domenic/sinon-chai) on the example above, to have more declarative assertions)

Also, and note that I'm not that familiar with sinon, maybe we could use the `sinon.replace` syntax, instead of needing to create a fake and a stub:
```js
beforeEach('set up stubs', function () {
  sinon.replace(add, 'default', sinon.fake())
});

afterEach('restore', function () {
  add.default.restore()
});

it('should call add script with an alias and a filename', function(done) {
  program.parse(['node', 'zos', 'add', 'ImplV1:Impl', '--skip-compile'])
  add.default.should.have.been.calledOnceWithExactly({ contractsData: [ { name: 'ImplV1', alias: 'Impl' } ] })
});
```

What do you think? We could have a single beforeEach block that stubs out all the `script`s we need for that test suite, plus `runWithTruffle`, and then each test should just be a two-liner with the arguments to the command, and the arguments to the stub. We could also make these blocks reusable across all `command` tests, but let's start with `add` for now.

bakaoh:
@spalladino -- actually i used `beforeEach`/`afterEach` and `sinon.replace` at first just like your snippet. But i can not find a way to wait for `program.parse` to finish execute command. Please let me know if you have any idea

spalladino:
I'd check if [commander](https://github.com/tj/commander.js/) has any event to notify when the execution of an `action` has completed, and try to use that. 

If not possible, then I'd try to abstract everything into a function that takes care of the boilerplate, so each test is as slim as possible. For example, something like:

```js
function itShouldParse(name, cmd, args, cb) {
  it(name, function (done) {
    let stub, fake;
    fake = sinon.fake(function () {
      cb(fake);
      stub.restore();
      done();
    })
    stub = sinon.stub(cmd, 'default').callsFake(fake)
    program.parse(args)
  });
}

itShouldParse('add with an alias and filename', add, ['node', 'zos', 'add', 'ImplV1:Impl', '--skip-compile'], function (fake) {
  fake.should.have.been.calledOnceWithExactly({ contractsData: [ { name: 'ImplV1', alias: 'Impl' } ])
});
```
I used the stub/fake syntax you presented originally, but it may work with replace as well. WDYT?

bakaoh:
your `itShouldParse` make each test so slim, but it's not flexible enough for tests that need more than 1 stub so i modified it a bit

test/commands/add.test.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+require('../setup')
+
+import {setup, itShouldParse} from './share';
+
+contract('add command', function() {
+
+  setup()
```
spalladino: I'd rename `setup` to `stubCommands`

test/commands/add.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+  itShouldParse('should call add script with an alias and a filename', 'add', ['node', 'zos', 'add', 'ImplV1:Impl', '--skip-compile'], function(fake) {
```
spalladino: WDYT about accepting a string with the command line input, instead of the already-split argv, just for ease of writeability in the tests? So, instead of `['node', 'zos', 'add', 'ImplV1:Impl', '--skip-compile']`, you could just write `node zos add ImplV1:impl --skip-compile`. This would only require a `split(' ')` in the `itShouldParse` function.

test/commands/share.js
```diff
@@ -0,0 +1,39 @@
...
+import * as push from '../../src/scripts/push';
+import * as runWithTruffle from '../../src/utils/runWithTruffle'
+import Session from '../../src/models/network/Session'
+import program from '../../src/bin/program';
+
+exports.setup = function () {
+  beforeEach('set up stubs', function () {
+    this.add = sinon.stub(add, 'default')
+    this.addAll = sinon.stub(addAll, 'default')
+    this.push = sinon.stub(push, 'default')
```
spalladino: Later, we'll need to stub all commands here. Note that you can access a list of all commands from `src/commands/index` and loop over it creating the stubs. However, I'm not sure if the stub will work correctly when imported from that list, but it may be worth a shot.

test/commands/add.test.js
bakaoh: i also prepended `node` so the input is exactly as usage `zos add ImplV1:impl --skip-compile`

test/commands/share.js
bakaoh: actually we stub `src/scripts/...` here so there're no `index` file. Even if we create an `index` for `scripts` we till need to change all `import` in `commands` to make it stubbale

bakaoh:
It's done. I have changed few things in `push` and `init` commands since i guess they're bugs, and make a change in `session` command to make it testable. Please preview them

src/commands/init.js
```diff
@@ -22,10 +22,10 @@ const register = program => program
 async function action(name, version, options) {
   const { force } = options
   if (options.lib) {
-    if (options.stdlib) throw Error('Cannot set a stdlib in a library project')
+    if (options.link) throw Error('Cannot set a stdlib in a library project')
     await initLib({ name, version, force })
   } else {
-    const { stdlib: stdlibNameVersion, install: installLib } = options
+    const { link: stdlibNameVersion, install: installLib } = options
```
spalladino: Whoa, good catch! Glad to see the tests are already paying off!

test/commands/share.js
```diff
@@ -0,0 +1,96 @@
...
+import * as remove from '../../src/scripts/remove';
+import * as session from '../../src/scripts/session';
+import * as status from '../../src/scripts/status';
+import * as update from '../../src/scripts/update';
+
+import * as runWithTruffle from '../../src/utils/runWithTruffle';
+import Session from '../../src/models/network/Session';
+import program from '../../src/bin/program';
+
+program.Command.prototype.parseReset = function() {
```
spalladino: I'm thinking if, instead of resetting the program, we could have a function to create a new one a reconfigure it on every test, so we don't need to depend on commander's internal structure. On the other hand, this approach works, is more concise, and also faster, so I guess it's the best option.

spalladino:
Merged! Thanks so much for your efforts, @bakaoh!

Issue #320, opened 2018-07-16
Logic contracts can be "accidentally killed"


spalladino:
Logic contracts registered in an app are currently open to the same attack as the [2nd Parity Wallet hack](https://blog.zeppelinos.org/parity-wallet-hack-reloaded/). Given that we one of our goals is to make deployments more secure, I'd like to address this at least before the anniversary of the hack (November 7th).

All jokes aside, we are not initializing logic contracts as they are deployed via `zos push`. This will also be true as we move onto the new _constructor contracts_ pattern. 

These are the options that come to mind to prevent the issue:

1- Push forward https://github.com/zeppelinos/zos-lib/issues/33, which prevents any calls whatsoever to a logic contract. This is the safest solution by far. It requires a new feature altogether in Solidity though, or meddling with the assembly code, which would break verifications, and potentially introduce unexpected bugs. However, even with the new feature, it requires the user to change their contracts to be used with zOS, something we're trying to avoid (unless we modify it automatically from the CLI before deployment).

2- Check the contract code for any SELFDESTRUCT calls. We could either check the binary (which would provide little info the user as where the call is issued, and may also fire some false positives if there is raw data with the same value as the opcode), or the source code of the contract and all its ancestors. The best we can do here is print a warning to the user, indicating that contract _could be_ insecure.

3- Force initialization of logic contracts when registered. This adds a lot of burden to the end user, who needs to provide "fake" params (which will never be used) for every contract to be registered. It also doesn't protect against this problem, since a SELFDESTRUCT in the contract logic could be callable regardless of the initialization parameteres.

I'd go with (2) for the time being. Thoughts?

frangio:
Agree on going with (2) for now.

Eventually we could use trustless off-chain computation :sparkles: (e.g. TrueBit) to perform all of these safety checks on contracts. I'm not sure if that would be in App, Directory, or elsewhere though...

nventuro:
Instead of looking at the raw bytecode (which will always contain raw binary data because of the [included metadata](https://solidity.readthedocs.io/en/v0.4.24/metadata.html)), and parsing the source code (which may be a bit fragile), we could traverse the AST (recursively), where a selfdestruct call will show up as an expression with the following fields: 

```
"typeDescriptions": {
  "typeIdentifier": "t_function_selfdestruct_nonpayable$_t_address_$returns$__$",
  "typeString": "function (address)"
}
```

[...]

PR #347, opened 2018-08-01
Warn if contract has a selfdestruct call

hardlydifficult:
Fix for #320

We are scanning the AST as suggested by @nventuro


hardlydifficult:
I'm working on the test fails now.

hardlydifficult:
@spalladino Correct, the original approach did not work. I had misunderstood the structure of the AST. I made the changes you suggested and added a Test. Please let me know if this looks okay.

I'm unsure how to address ancestor concern.  I wrote a test for this, and commented it out because it fails ATM.  If you have a suggestion on how to proceed, I'm happy to take a crack at it.

test/scripts/add.test.js
```diff
@@ -92,10 +92,26 @@ contract('add script', function() {
       this.logs.errors[0].should.match(/constructor/i);
     });
 
-    it('should not warn when adding a contract without a constructor', async function() {
+    it('should not warn when adding a contract without a constructor or a selfdestruct cull', async function() {
```
spalladino: typo: cull => call

src/models/local/LocalBaseController.js
```diff
@@ -92,6 +97,26 @@ export default class LocalBaseController {
...
 
+  hasSelfDestruct(contractDataPath) {
+    if (!fs.exists(contractDataPath)) return false
+    const ast = fs.parseJson(contractDataPath).ast
+    return this.hasKeyValue(ast, "typeIdentifier", "t_function_selfdestruct_nonpayable$_t_address_$returns$__$");
+  }
+
+  hasKeyValue(data, key, value) {
+    if (!data) return false
+    if (data[key] === value) return true
```
spalladino: Note that at this point you're already assuming that `data` is an object, making the check below redundant

src/models/local/LocalBaseController.js
```diff
@@ -92,6 +97,26 @@ export default class LocalBaseController {
...
+    if (typeof(data) === 'object') {
+      for (const child_key in data) {
+        const child_value = data[child_key];
+        if (this.hasKeyValue(child_value, key, value)) {
```
spalladino: I think the type check should be moved onto here: only call `this.hasKeyValue` if `child_value` is an object.

src/models/local/LocalBaseController.js
```diff
@@ -92,6 +97,26 @@ export default class LocalBaseController {
...
+    if (!fs.exists(contractDataPath)) return false
+    const ast = fs.parseJson(contractDataPath).ast
+    return this.hasKeyValue(ast, "typeIdentifier", "t_function_selfdestruct_nonpayable$_t_address_$returns$__$");
+  }
+
+  hasKeyValue(data, key, value) {
+    if (!data) return false
+    if (data[key] === value) return true
+    if (typeof(data) === 'object') {
+      for (const child_key in data) {
```
spalladino: Please make sure to use `camelCase` for all variable names, and not `snake_case`, for consistency

test/scripts/add.test.js
```diff
@@ -92,10 +92,26 @@ contract('add script', function() {
...
+    
+    it('should warn when adding a contract with a selfdestruct call', async function() {
+      add({ contractsData: [{ name: 'WithSelfDestruct', alias: 'WithSelfDestruct' }], packageFile: this.packageFile });
+
+      this.logs.warns.should.have.lengthOf(1);
+      this.logs.warns[0].should.match(/selfdestruct/i);
+    });
+    
+    // TODO: This warning is not yet implemented
+    // it('should warn when adding a contract who\'s parent has a selfdestruct call', async function() {
```
spalladino: Use `it.skip` instead of commenting out the code if it's still unimplemented

src/models/local/LocalBaseController.js
```diff
@@ -92,6 +97,27 @@ export default class LocalBaseController {
     return !!abi.find(fn => fn.type === "constructor");
   }
 
+  hasSelfDestruct(contractDataPath) {
+    if (!fs.exists(contractDataPath)) return false
+    const ast = fs.parseJson(contractDataPath).ast
+    if (this.hasKeyValue(ast, "typeIdentifier", "t_function_selfdestruct_nonpayable$_t_address_$returns$__$")) return true
+    for (let i = 0; i < ast.nodes.length; i++) {
+      for (let j = 0; j < (ast.nodes.baseContracts || []).length; j++) {
```
spalladino: Could you change these for-loops to use either the [`for-of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) construct for [`forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) method, for better readability?

hardlydifficult:
RE delegate call: happy too.  I looked at this a bit but couldn't figure out how to use delegate call (never tried this in a contract before). Could you provide am example I can test with?  I tried the following but it does not work (no error, but it does not actually selfdestruct the contract):

this.delegatecall(bytes4(keccak256("selfdestruct(uint256)")), msg.sender);

thanks,
nick.

spalladino:
A `delegatecall` is written as a regular low-level `call`, the example you provided should work, but you'll probably need to be using a target address different than `this` (see [here](https://gist.github.com/sogoiii/f0ced0a4e569b5f38d302e7072d78b43) for an example). The goal would be to see how a delegatecall looks like in the AST, and capture it, as you did with the selfdestruct. Thanks!!

hardlydifficult:
Delegate call does not seem to create a clear entry in the AST we could scan for.  A couple options I can think of:

 - Look for: 'typeIdentifier' == "t_stringliteral_ecd652fcce345f51df38ced4d0016cc21c8a20495863986a365f73a38368985d"
 - Or look for "t_contract$_Delegate_$42" and provide a more generic warning against making low level calls.

Either of these seem okay?

spalladino:
@hardlydifficult I made a small example, and the AST seems to record the delegate call with the type `t_function_baredelegatecall_nonpayable$__$returns$_t_bool_$`. 

@nventuro or @ajsantander, given that you are working with the Solidity compiler internals, could you confirm if this is always the case for a delegatecall (at least when not using assembly)?

hardlydifficult:
Quick update: I have the stack overflow fixed locally, but currently investigating another issue.  

(the AST is referencing every class in the .sol file, vs just ones related to the contract being tested... leading to false positives)

Will continue tomorrow.

hardlydifficult:
I believe all the outstanding feedback has been addressed.  Let me know if there's anything I missed.

thanks,
nick.

spalladino:
Awesome @hardlydifficult, thanks for the hard work!! I'll port the commit to the new monorepo at zeppelinos/zos.

